PR #4, opened 2017-01-20
Add Qt to the RPATH

jacky405:
Add Qt to the RPATH, so that there is no need to set LD_LIBRARY_PATH at
runtime if Qt is installed in a non-standard location


GeniviDLTmaintainer:
Pushed changes after successful Jenkins build on Ubuntu 16.04 from local repo

PR #5, opened 2017-02-10
Fix build on OS X, support HiDPI displays

neverpanic:
These patches fix the build of current dlt-viewer on macOS and improve rendering on HiDPI screens such as retina displays.


GeniviDLTmaintainer:
Just pushed changes from BMW git ...

PR #7, opened 2017-08-31
Start ColorDialog with current color instead of white

mikelima:
This patch makes the color selector start with the current filter color to modify in the color picker dialog instead of the default white color. It makes it easier to adjust the filter color if it is too similar to one already used.


GeniviDLTmaintainer:
Come with next pachtes ...

GeniviDLTmaintainer:
Came with push on 5.1.17

PR #8, opened 2017-08-31
Fix huge window problem whith long filenames

mikelima:
When the filename/path was too long, the statusbar width becomes very
long, and the whole window becomes very wide as well. This patch elides
text from the variable width fields of the statusbar to avoid this. The
full text of these fields is available as a tooltip.


GeniviDLTmaintainer:
comes with next push from local master

PR #9, opened 2017-08-31
Use a combo widget for the search toolbar

mikelima:
This allows to switch between the last few search expressions.
This is convenient at times when one wants to search alternatively for more than one pattern.


GeniviDLTmaintainer:
Will come with next push from local repo

PR #10, opened 2017-12-23
Fix duplicate symbols

roland-wolf:
On Linux buttons in file transfer plugin did not work. This was caused by duplicate symbols in plugins. 
Form classes were put in namespaces to avoid this situation. 

First commit changes build system to make plugins work out of the box in build directory. All plugins will now be built in a plugins directory.


GeniviDLTmaintainer:
Comes with next push from local master

PR #12, opened 2018-03-08
Add CMake build support under windows

zulkarnine:
Fix a common problem with QTDIR
Fix compile and link errors while building the project in windows using CMake and (VS2013)


CMakeLists.txt
```diff
@@ -20,6 +20,8 @@ project (DLT-Viewer)
 
 set(CMAKE_CXX_STANDARD 11)
 
+set(CMAKE_PREFIX_PATH $ENV{QTDIR})
```
tobsan: Shouldn't this be `set(CMAKE_PREFIX_PATH "${CMAKE_PREFIX_PATH} $ENV{QTDIR}")` instead, to avoid overwriting existing prefix path?

zulkarnine: Yes, that would be safer. I will change it

GeniviDLTmaintainer:
Comes with next push from local muster ...

zulkarnine:
> Comes with next push from local muster ...
What does it mean?

PR #13, opened 2018-03-10
Port dlt-viewer to QSerialPort

jjardon:
No need to use qextserialport anymore; QSerialPort Class (http://doc.qt.io/qt-5/qserialport.html) is now part of Qt

Not tested: windows build


gunnarx:
- I assume 2b715d8 won't compile on its own, so you should squash the two commits into one, IMHO.
- What is the minimum version of Qt that is required to have QSerialPort available?
-- ... and is that dependency documented -- the README still says it needs Qt 5.5.1, which I suspect might not be enough?

(Note that I'm not the maintainer, just reviewing...)

jjardon:
- It will compile if you apply the commits in the order are presented in this branch: first you port to QSerialPort, then you remove the external library as It's not needed anymore
- QSerialPort is there since Qt 5.1 (http://doc.qt.io/qt-5/qserialport.html)

gunnarx:
Oh, GitHub is doing some weirdness with "new changes since last you viewed" (which won't go away even though I have viewed both commits!)  So it confused me on the listed commit order.  All good then, and that's a nice PR, and seems it is a long time coming if this has been the case since Qt 5.1(!)

AWWenzel:
Thanks for the great contribution. We did not removed the external qextserialport library in the past, because we wanted to support also older LTS Linux distributions. But now i think it is the right time.
On which platform (Linux, Windows) did you test it?
We will test the patch and then merge it into the master.

jjardon:
@AWWenzel Hi, I have tested it in Arch Linux, with both the cmake and qmake build system (only build in the later)

Let me know if you want me to test something specifically

GeniviDLTmaintainer:
Will come with next push from local master, had to do some changes because we had some local fixes here inbetween. Also had to put in comment the line 
ecuitem->m_serialport->waitForReadyRead(10);
as it does not work with Windows and Ubuntu

jjardon:
@GeniviAMmaintainer mmm, this is rebased against current master and it seems to build fine (both with qmake and cmake)

About the crash; it was working ok in the testing I did in my Arch system, but maybe I have missed something when rebasing as this was done a while ago; I will recheck

GeniviDLTmaintainer:
Just came with latest push from local master

PR #15, opened 2018-05-11
Fix file copy error in SDK generation batch file

ysk-sato:
Error in SDK batch file due to missing "README.txt" is fixed.
And also non-existing file name is removed from qt project file.
(This is only linked with project expplorer in Qt creator)

Signed-off-by: Yusuke Sato <yusuke-sato@apn.alpine.co.jp>


GeniviDLTmaintainer:
Just came with latest changes from local master

PR #17, opened 2018-06-13
Fixing unstable behaviour in MainWindow::nearest_line

WagnerMi:
The implementation of MainWindow::nearest_line should search for the
the line of a given index value.

If the index is present in the (filtered)message list the according
line is returned. If it is not present the next present index smaller
than the searched one is used.

The method has been working instable.
The reason for this is the qfile.getIndexFilter() is expected to
return a sorted list. This is not the case.
If the returned list is not sorted, wrong results are returned.

The new version sorts the list and searches in this sorted list.


GeniviDLTmaintainer:
Will come with next push to master

PR #18, opened 2018-06-28
Add "Resize columns to fit" and "Copy Selection Payload to Clipboard" context menu items to main and search tables

eBardieCT:
The attached patch adds a "Resize columns to fit" item to the main table's context menu.

### DIFF
--- a/src/dltexporter.cpp
+++ b/src/dltexporter.cpp
@@ -158,7 +158,8 @@ bool DltExporter::finish()
         /* close output file */
         to->close();
     }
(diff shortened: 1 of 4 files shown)


eBardieCT:
Hmmn. Apologies for the inadequate commit messages.

The "Resize columns to fit" commits add a context menu item to each of the tables (main and search) which does what it says on the tin.

The "Copy Selection Payload to Clipboard" commit adds a context menu item to each of the tables which copies only the Payload field of each selected DLT entry to the Clipboard.

GeniviDLTmaintainer:
Came with last push

PR #21, opened 2018-10-11
"Copy Selection Payload" menu item

eBardieCT:
Sometimes when debugging a problem all I want is the DLT entry's payload,  e.g. to use when grepping the source.

Rather than copying the entire entry and then editing this down, this patch adds a context menu, to both tables, to copy just the payload.

It also adds a context menu to the search table to implement the 'Resize to fit' option recently added to the main table.


GeniviDLTmaintainer:
Came with latest push

PR #22, opened 2018-12-11
Ebardie/filter on appid regex

eBardieCT:
Add regular expression option to Application Id field in filters. This can save on the tedious process of creating/managing multiple filters.  Balks at the field containing more than four chars in non-regex mode.


GeniviDLTmaintainer:
Came with latest push

PR #29, opened 2019-05-08
Fix the manual trigger to re-read data after reapplying filters

vrupanov:
This should fix the following issue: https://github.com/GENIVI/dlt-viewer/issues/28

Has been tested only manually, and might be that it only reduces the scope of the problem.


alexmucde:
Verfified on our CI and pushed to master.

PR #31, opened 2019-06-06
QTextStream wants an IO target.

gdelmondo:
Fixes a crash when a plugin failes to load


GeniviDLTmaintainer:
Hi ! How can we provoke the crash without this fix ? What kind of problem when loading a plugin ?

gdelmondo:
> Hi ! How can we provoke the crash without this fix ? What kind of problem when loading a plugin ?

Plugins can have additional dependencies the main application does not have. If one of these dependency is not met, the plugin fails to load.

For example, if you remove the Qt5PrintSupport.dll from the library path, the filetransferplugin fails to load and triggers this bug.
With the patch, dlt-viewer properly handles this situation.

alexmucde:
Verfified on our CI and pushed to master.

PR #32, opened 2019-06-07
fixed DltTableView display bug

enoy19:
The payload (and actually all cells) started wrapping words after I upgraded my system to Fedora 30.
Since I don't have much Qt experience, I did a quick fix to read the payload better.
This fix just removed the vertical center alignment of payload cells in the TableModel.

Example: the payload text is wrapped to 3 lines. I can only view the second line because the cells height wasn't big enough to fit all 3 lines.

This is just a quick and dirty fix for my situation. I hope it's still helpful to find the root cause.
Perhaps this is a system specific issue. If there is any setting (or Qt version) I need to use, please tell me :)

For the long term it would be better to disable the word wrap completely.

### DIFF
--- a/src/tablemodel.cpp
+++ b/src/tablemodel.cpp
@@ -388,7 +388,7 @@ TableModel::TableModel(const QString & /*data*/, QObject *parent)
             case FieldNames::ArgCount:
                 return QVariant(Qt::AlignRight  | Qt::AlignVCenter);
             case FieldNames::Payload:
-                return QVariant(Qt::AlignLeft   | Qt::AlignVCenter);
+                return QVariant(Qt::AlignLeft);
         }
     }
 


enoy19:
I just noticed, that the searchtablemodel has the same display issue.

AWWenzel:
@enoy19 I tried this patch under Windows Qt 5.13.0, then the vertical alignment does not work anymore. Perhaps there is anothe issue, which has nothing to do directly with the alignment?
Do you have a screenshot?

AWWenzel:
@enoy19 Please also retest lattest version on github. Perhaps this already solves your problem.

GeniviDLTmaintainer:
Fix will cause other problems and cannot be reproduced. If you have another solution please reopen ticket.

enoy19:
@GeniviDLTmaintainer  I confirm, that this bug does not exist anymore in the latest version. Might have been a qt+gnome bug on my machine.

PR #73, opened 2019-11-22
Ignoring *.orig backup files from KDiff3

olaf-dreyer:
Signed-off-by: Olaf Dreyer <olaf.dreyer@partner.magna.com>


alexmucde:
All builds succesful.

PR #76, opened 2019-11-25
Fix compiling issues

olaf-dreyer:


alexmucde:
You have removed the console option of the Qt build, but we use the DLT Viewer also in a console version. Does the command line functions still work with this build? Did you have tested this?

@olaf-dreyer there is a conflict now, because of another pull request. Can you fix this please?

alexmucde:
Applied patch manually on master.

alexmucde:
Tested this version without console option. Command line not working anymore. Added console again in a manually added patch to master.

PR #78, opened 2019-11-25
Fix non verbose plugin

olaf-dreyer:


alexmucde:
@olaf-dreyer Here the same. The first three commits must be removed.

alexmucde:
Applied patch manually on master.

PR #82, opened 2019-11-28
Fix filter item is checked but not enabled

olaf-dreyer:
My patch for the multiselection mode of the filter tree view isn't working correctly,
since I forgot to enable the filters and not just set the check state.


alexmucde:
@olaf-dreyer you have again added a merge commit to the pull request. I will apply this patch manually. Next time please provide a pull request without any merge commit.

alexmucde:
Manually integrated and pushed to master

PR #96, opened 2020-03-30
Fix: macOS qmake build

Morheit:
Corrected the `rpath` option for macOS

The macOS uses the Clang compiler. The [`rpath` syntax for Clang](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-rpath) differs from GCC.

- Clang: -rpath <arg>
- GCC: -rpath=dir

The change was introduced due to the build failing with error:
```
ld: unknown option: -rpath=.
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

### DIFF
--- a/src/src.pro
+++ b/src/src.pro
@@ -20,7 +20,11 @@ QT_VER_MIN = $$member(QT_VERSION, 1)
 
 CONFIG += console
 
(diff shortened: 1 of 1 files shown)


Morheit:
Hello @alexmucde,

Could you review the pull request, please?

Thank you!

PR #102, opened 2020-06-02
Fix case insensitive regexp match in filters

bazurbat:
- Fix inverted condition when setting pattern options.
- Fix incorrect passing of case sensitivity flag as match offset which prevented the ^ anchor from working.


alexmucde:

`Signed-off-by: Alexander Wenzel <Alexander.AW.Wenzel@bmw.de>`

PR #104, opened 2020-06-12
restore settings option ShowArguments

johumuc:
Restored Setting option "Show Argument Columns".  This was removed by Gernot Wirschal in 
b438bef7e0aa1a61cd39f342a7ba511540951f52 in 2018 but however the real feature itself was not removed cleanly and is/was still working. Only the Settings UI was missing today. 
It seems the Feature was not self explaining.
I added some sentences to the user manual tex file.

Scope of the feature:
instead of reading the squashed PAYLOAD text representation, you are able to see the single arguments split to a separate column called Arg0, Arg1,...ArgX.
The Columns are placed at the right end of the tableview, so you will have to scroll far right if you use it the first time. 
Advantage: 
If the first Argument of your DLT Messages  is e.g. a dynamic text but  you want to  compare the following Arguments , it is often hard to spot as the the arguments are not aligned  due to the variable length of the first/prior arguments.

e.g. 
```
|Payload                                                                 |
|test.c:Line1 data 43 05 47 54 39 59 43 58                               |
|myverylongservicename.c:Line154543634653465 data 43 05 47 54 38 59 43 58|
```

By splitting the arguments in separate columns they are perfectly aligned and differences can be spotted more easily.

```
| Arg0                   | Arg1                    | Arg3 | Arg4                   |
|test.c                  |Line1                    | data |43 05 47 54 39 59 43 58 |
|myverylongservicename.c |Line154543634653465      | data |43 05 47 54 38 59 43 58 |
```


alexmucde:

alexmucde:
Merged manually.

PR #105, opened 2020-06-14
Extended non-verbose messageid support (column,filter,search)

johumuc:
Enhancements of non-verbose  messageID Support 
User Feature:
-optional separate column for Message ID
-customizable format for Message ID Column (sprintf notation)
-Filter option for Message ID  with hex or dec input
-Filter option for Message ID Range  with  hex or dec input
-enable search for MessageID in custom format in header data, 
  even when decoded with plugin

Code:
simplified column handling by
-moving duplicate code for columns handling to fieldnames.cpp
-reduced inconsistencies in column settings for search/main table
-new column only needs changes in fieldnames.cpp now
-no more updating of column count defines need for new columns


alexmucde:

alexmucde:
Merged manually.

PR #120, opened 2020-10-21
Fixed scroll in table view in case when Index column is hidden

AndreySakun:
Fixed scroll in table view in case when Index column is hidden; 
Disabled autoscroll to the right in search results

Signed-off-by: Andrey Sakun sakun.andrey@gmail.com


alexmucde:
@AndreySakun Can you update your commit message with signed off statement?

alexmucde:
I tested the PR and the bug fixies works fine. But this did not solve the issue when index is disabled, the marked line will not marked again when filters are disabled. New fix needed for this.

alexmucde (approved):
Verified changes on my own build  and it works.

AndreySakun:
> @AndreySakun Can you update your commit message with signed off statement?

Done

Issue #115, opened 2020-08-31
Issue displaying DLT message containing other DLT messages


andreirusu96:
Hello!

I am currently using the Filetransfer feature of Genivi DLT to transfer a buffer containing a bunch of DLT logs (complete DLT logs containing the "DLT0x01" header). As you know, the Filetransfer basically sends DLT logs to DLT Viewer with the contents of the file as DLT_RAW(rawData, dataSize). However, there is an issue inside DLT Viewer that causes the DLT Filetransfer log (the "FLDA ... FLDA" one) to show as "CORRUPTED MESSAGE!!!" when it contains more than 2 DLT logs within itself. Please see below screenshot:

![image](https://user-images.githubusercontent.com/16763553/91775003-bd5fb300-ebf2-11ea-9568-16fe3984792c.png)

As you can see, the messages inside the payload are being read as actual DLT messages. How it should show up:

![image](https://user-images.githubusercontent.com/16763553/91775114-f9931380-ebf2-11ea-8946-5b7a0cabe832.png)

**Observations from testing:**

- At least 2 logs need to be contained within the "FLDA ... FLDA" Log for the corruption to happen (haven't managed to check out why this issue doesn't occur only one DLT message inside the payload)
- There was this commit that fixed receiving messages containing the DLT header in the payload ("DLT0x01") : https://github.com/GENIVI/dlt-viewer/commit/40e94c58bf7e21cf24d07d0927d46317a4aa3723 This did indeed fix the issue, however, this specific edge case was missed (see below).

**Observations from debugging:**

1. The "CORRUPTED MESSAGE!!!" error appears only when the "FLDA ... FLDA" log is the last log read in the "buf" variable:  
`buf = files[numFile]->infile.read(READ_BUF_SZ);`(line 204, file "qdltfile.cpp")

- In this case, the following would be true at the end of the method call: `file_size == next_message_pos` , so the following method will be called: `files[numFile]->indexAll.append(current_message_pos);` (line 232, file "qdltfile.cpp")
- The aforementioned commit fixed the handling of such messages when the "FLDA ... FLDA" would be placed in begging/middle of the "buf" buffer, but this specific case was missed. 
[...]

PR #121, opened 2020-10-22
Fix read of DLT with DLT header in payload

andreirusu96:
Please check Issue #115 for more details.

Signed-off-by: Andrei Rusu <andrew_x_2005@yahoo.com>


alexmucde:
Can you provide a DLT File which contains a DLT message header in the payload?
Can you describe the use case when the problem occurs?

alexmucde (approved):

PR #123, opened 2020-11-05
Set scaling attributes after create QApplication

m42e:
Qt requires these attributes set after the creation else it will stay
blurry on macOS.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -26,12 +26,13 @@
 
 int main(int argc, char *argv[])
 {
+
+    QApplication a(argc, argv);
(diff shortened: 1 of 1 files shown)


OXINARF:
This fixes it for Windows too, although some parts of the interface get quite big:

image

Notice the main table headers and scrollbar. Also the icons on the toolbar are way too big.

alexmucde (approved):
Looks good on my PC with FullHD and also 4k. Tried it on Windows with Qt 5.12.9 and MSVC2015.

PR #131, opened 2020-11-13
Sort by time/timestamp keeps index order

### LINKED ISSUE #124: Enabling "Sorted by Time" or "Sorted by Timestamp" sorts logs in wrong order
image
I have this issue that whenever I apply a filter and for any group of rows there is same Time then rows will be sorted in descending order (in image can be seen that Index and Count are also in descending order).
Even if I try to sort by timestamp (which is not always possible if I have multiple life cycles in same dlt) if time is same rows will be in descending order
image

I tried to find some settings that control this or something in source code itself, but didn't manage to fix this.

OXINARF:
For messages that have the exact same time/timestamp, we keep the same order for the messages by using their index.

Fixes #124

**This has not been tested**. @alexmucde if you can help with a test build it would be great.

### DIFF
--- a/src/dltfileindexer.cpp
+++ b/src/dltfileindexer.cpp
@@ -15,11 +15,20 @@ extern "C" {
     #include "dlt_user.h"
 }
 
-DltFileIndexerKey::DltFileIndexerKey(time_t time,unsigned int microseconds,unsigned int timestamp)
+DltFileIndexerKey::DltFileIndexerKey(time_t time, unsigned int microseconds, int index)
+    : timestamp(0)
 {
     this->time = time;
     this->microseconds = microseconds;
+    this->index = index;
+}
+
+DltFileIndexerKey::DltFileIndexerKey(unsigned int timestamp, int index)
+    : time(0)
+    , microseconds(0)
+{
     this->timestamp = timestamp;
+    this->index = index;
 }
 
 DltFileIndexer::DltFileIndexer(QObject *parent) :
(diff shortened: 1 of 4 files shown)


m42e:
Test Build for mac is what I could provide :)

m42e:
I can confirm it seems to work. no other tests performed.

OXINARF:
@m42e Thank you for testing! mac is unfortunately the only platform I don't have 🙂

OXINARF:
I've just realized that there is an AppVeyor build for Windows that has a build artifact (https://ci.appveyor.com/project/alexmucde/dlt-viewer/builds/36299389/artifacts). I've just tried it and it seems to be working OK.

@alexmucde It would be good if Windows AppVeyor builds were also part of the GitHub checks, it could be failing on this PR and I wouldn't know.

alexmucde:
AppVeyor will build for Windows and Ubuntu platform, but it looks like the approval is overwritten by the last build. Perhaps we can get it all in one build.

alexmucde:
@OXINARF Can you please update you commit message with a signed off statement?

alexmucde (approved):

alexmucde:
Tested on Windows and it works, verified it with old version ;-)

OXINARF:
> @OXINARF Can you please update you commit message with a signed off statement?

Done

> AppVeyor will build for Windows and Ubuntu platform, but it looks like the approval is overwritten by the last build. Perhaps we can get it all in one build.

Yeah, that would be nice. Or change to GitHub Actions. I might look into that when my schedule isn't so full 🙂

PR #132, opened 2020-11-13
Load dlf from subfolders

### LINKED ISSUE #126: Allow subdirectories in default filter folder
Allow subdirectories in default filter folder to organize filters.

m42e:
Enabled the loading of .dlf from subfolders within the default filter
location. Symlinks are also allowed.

Fixes #126


alexmucde (approved):
Tested under Windows with Qt  5.12.9 and MSVC2015.

PR #133, opened 2020-11-13
Enable optional append for default filters

### LINKED ISSUE #128: Allow Append from the default filter folder
Appending of a filter from the default filter folder would be useful.

m42e:
If the checkbox is activated and default filter is selected it is
appended rather than replacing the previous filters.

Fixes #128

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -510,7 +510,6 @@ void MainWindow::initFileHandling()
     ui->checkBoxSortByTimestamp->setEnabled(ui->filtersEnabled->isChecked());
     ui->checkBoxSortByTimestamp->setChecked(QDltSettingsManager::getInstance()->value("startup/sortByTimestampEnabled", false).toBool());
 
-
     /* Process Project */
     if(QDltOptManager::getInstance()->isProjectFile())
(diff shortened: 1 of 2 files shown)


alexmucde:
I tried this new feature, but it was hard for me at the beginning to underdstand, where the "Append Filter" checkbox belongs to and what it really does. After some tries i understood that instead of switching the filter you will append the selected filter configuration to the current filter.

The original idea of default filter was that you the filter index are stored in the cache , which are generated in parallel and you can quickly switch between different filter configuration. This  new feature will not be able to switch quickly between filter configurations, when you append the current filter.

Perhaps you can move the checkbox beside or under the Default Filter list, that the user knows that it belongs to this feature.

m42e:
Ok i thought this looks a bit messy, too. Maybe I put it in a line just below the selectbox

alexmucde (approved):
Looks good now, tested with Qt 5.12.9 and MSVC2015 under Windows.

alexmucde:
Github cannot merge. Please can you manually rebase to master?

alexmucde:
Merge worked, will check if it also compiles.

PR #134, opened 2020-11-15
Fix filter load in case of error

m42e:
If a corrupted file has been passed to LoadFilter it still tried to load
the filters after notifying the loading of the file failed. This caused
a crash. Also the duplicate Messagebox has been removed.

Signed-off-by: Matthias Bilger <matthias@bilger.info>


alexmucde (approved):
Tested with Qt 5.12.9 in Windows with MSVC2015.

Issue #127, opened 2020-11-12
Allow search in default filter folder


m42e:
Add search for a filter in the default filter folder select list.

alexmucde:
A pull request also for this feature would be great.

alexmucde:
#135 integrated

PR #135, opened 2020-11-15
Combo Box for default filter is capable of search

m42e:
Allow to search a filter by a piece of the name in the defaultfilter
combobox.

Fixeds #127


alexmucde (approved):
Tested with Qt 5.12.9 on Windows with MSVC2015.

PR #136, opened 2020-11-15
Using better threshold for text color

m42e:
Improved the selection of the text color based on the background. Also
fixed calculation (the real value has been used for red rather than the
uint8 value).

### DIFF
--- a/src/dltuiutils.cpp
+++ b/src/dltuiutils.cpp
@@ -13,8 +13,6 @@ QColor DltUiUtils::optimalTextColor(QColor forColor)
     // Counting the perceptive luminance - human eye favors green color...
     int a = ( 0.299 * forColor.red() + 0.587 * forColor.green() + 0.114 * forColor.blue());
 
-    qCritical() << forColor <<" " <<  a;
-
     if (a > 166)
     {
         d = 0; // bright colors - black font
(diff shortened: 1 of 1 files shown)


src/dltuiutils.cpp
```diff
@@ -11,9 +11,11 @@ QColor DltUiUtils::optimalTextColor(QColor forColor)
     int d = 0;
 
     // Counting the perceptive luminance - human eye favors green color...
-    double a = 1 - ( 0.299 * forColor.redF() + 0.587 * forColor.green() + 0.114 * forColor.blue())/255;
+    int a = ( 0.299 * forColor.red() + 0.587 * forColor.green() + 0.114 * forColor.blue());
 
-    if (a < 0.5)
+    qCritical() << forColor <<" " <<  a;
```
alexmucde: There is an build error in Windows:

./../src/dltuiutils.cpp:16:15: error: invalid use of incomplete type ‘class QDebug’

qCritical should also not be called everytime, only in error situations.

alexmucde (changes requested):

src/dltuiutils.cpp
m42e: This has already been fixed, i do not see this anymore

alexmucde (approved):

alexmucde (approved):
Tested with Qt5.12.19 and MSVC2015 under Windows

PR #137, opened 2020-11-15
Export compile commands

m42e:


m42e:
@alexmucde Is there a possibility to retrigger the build job? There seems to have happened a timeout.

PR #146, opened 2020-11-26
Use QPalette::Window instead deprecated QPalette::Background

jjardon:


alexmucde:
Please add a signed off statement to the commit message, then i can merge.

PR #152, opened 2020-12-02
Remove wrong (swapped) clone commands for macOS

hannesa2:
I recommend not to explain git fundamental basics, btw the commands are wrong, they should be swapped.
Conclusion: it's just noise


alexmucde (changes requested):
Signed off statement missing

hannesa2:
Signed off is done

alexmucde (approved):

PR #165, opened 2020-12-19
Build on GitHub

hannesa2:
This pull request builds macOS and Linux, like current Travis, but much faster and it keeps artifacts.

The missing Windows is a WIP https://github.com/hannesa2/dlt-viewer/pull/5 but I handle it as an extra pull request. 
But to be honest I've no Windows and can't verify


hannesa2:
Hm, as long as it's only on my fork it builds here nothing. Even Travis run, but I removed it.

Here https://github.com/hannesa2/dlt-viewer/pull/4 you can see how does it looks like

alexmucde:
This looks very good. Do you think it is ready for merging or do you still want to fix something?

hannesa2:
In my point of view, it's perfect.

lukx:
@alexmucde the Windows Build bundles some of the QT dlls, that's all dlls which are being copied to the install folder in the three *.bat-Files.
I have only tonight thought that one might need to check the Qt Licensing whether this affects the decision to bundle, or whether this project needs an update to its license declaration
Do you have an expert at hand to judge on that?

alexmucde:
Qt is distibuted under Open Source license and Commercial license. The Open Source license and the commercial license are not allowed to be mixed in a project. As we are an open source project we are allowed to use the open source distibution of Qt. Until a specific version Qt Company provided an open source build for Qt for Windows. For later versions only Source Code is provided. We are using currently the last version of Qt distributed directly by the Qt company, before the license change. Afterwards the Qt dlls must be build from source. I do not know, if the Qt distribution provided in the Github Action is selfbuild or provided directly from Qt company.

lukx:
According to the library I used, it's using (L)GPL - https://github.com/miurahr/aqtinstall/#license-and-copyright

hannesa2:
Is there something changes now compared to Travis build ?

build_sdk_windows_qt5_MSVC.bat
```diff
@@ -288,18 +288,13 @@ echo ####################################
 echo ###       ERROR occured          ###
 echo ####################################
 cd ..
-if '%WORKSPACE%'=='' (
-pause
-)
-rem exit 1
```
alexmucde, line 294: Can we add here, like in the DLT Parser script?

set /p name= Continue

Then i can run it also on my desktop and see the result.

build_sdk_windows_qt5_MSVC.bat
```diff
@@ -288,18 +288,13 @@ echo ####################################
...
 :QUIT
 echo ************************************
 echo ***       SUCCESS finish         ***
 echo ************************************
 cd ..
 echo SDK installed in: %DLT_VIEWER_SDK_DIR%
-if '%WORKSPACE%'=='' (
-pause
-)
-rem exit 0
```
alexmucde, line 305: Can we add here, like in the DLT Parser script?

set /p name= Continue

Then i can run it also on my desktop and see the result.

alexmucde (changes requested):
I tried the batch scripts on my desktop and they still work, except that the windows are closed, before i can see the results, in the DLT Viewer Build Script, see my comments. In the dlt Parser script it still works.

The qwt build script we can even remove. The DLT Viewer open source release has no plugins which are using the qwt library. Perhaps later we can add it again, if there is a need for it.

The Ubuntu 16.04 target can also be removed.

Can you change this? Then i will merge.

hannesa2:
I at least don't hear the call of duty for the Windows stuff. 
Can we make the requested changes on an additional pull request "Be able to build desktop windows", to move forward with this topic ?

Keep in mind: it builds, it works and only local issues are there

alexmucde:
I will merge now and then we can fix further things later.

hannesa2:
Ahhh, now it build here as well !

alexmucde:
Cool everything was succesful build, i will merge.

alexmucde:
Thanks very much for your help!

lukx:
Thank you so much for bringing this on :-) Maybe, in a future, we can even find the binaries on the github releases page

hannesa2:
> Maybe, in a future, we can even find the binaries on the github releases page

I already open an issue for that #172 but I need some help for this

PR #168, opened 2021-01-20
Raw Logging

alexmucde:
Raw Logging stores all received data from all ECUs in a single raw logging file. The raw logging file must be later converted into DLT format. This new feature reduces CPU consumption during logging about  five times. Especially on slow PCs this is important.


alexmucde:
Current version is only used for performance analysies, should not be merged yet.

The following things are missing:

- provide ECU/Project configuration with each raw log file
- provide connect/disconnect information in the raw log file
- converter into dlt file

hannesa2 (approved):
For a good review, it needs to be testes and build. I even don't know which IDE you use. -Sorry-

My opinion: As long as it builds you could merge it.

hannesa2:
Why this is green here 
image

and in #184 it's red. Who knows ?

Maybe someone can remove it. I don't know how

alexmucde:
Will not follow up further with this enhancement, maybe later with new PR. Will close now.

PR #169, opened 2021-01-22
Pause build windows on desktop when finished

alexmucde:


lukx:
I had assumed that .bat files would be executed from an interactive environment (a command-line), so that the window won't close anyway... 
It seems to work on CI like this, though.

lukx (approved):

alexmucde:
If you start it from command line window will not be closed, but if you start it from Explorer, windows will close.

PR #192, opened 2021-03-27
Disable RPATH usage:

LocutusOfBorg:
Make the usage of RPATH settings to detect non-standard QT
 installations optional.


alexmucde (approved):
Ok

PR #194, opened 2021-03-31
Remove using default filter index

alexmucde:
Signed-off-by: Alexander Wenzel Alexander.AW.Wenzel@bmw.de


alexmucde:
Switching between different filters in the default filter list was not showing anymore the correct filter index.
Removed the feature to use the default filter index, because it was not working anymore.
For fast loading of index the file index cache will be used instead.
In future a new feature can be added to create index filter cache files for default filter in background automatically.

PR #210, opened 2021-05-17
Keep whitespaces in ascii export

alexmucde:


jr45 (approved):
works as indicated

PR #225, opened 2021-10-28
User experience enhancements for filters

jschmer-garmin:
* add "Filter Clear all" option to Filter table context menu
* support deleting all selected filters (multi-selection support)
* support deleting all selected filters with the DELETE key


jschmer-garmin:
replaced by PR #227

PR #226, opened 2021-11-03
Windows build script improvements

jschmer-garmin:
Apparently, there are problems setting the errorlevel
variable after rmdir was called and if errorlevel is
checked afterwards it will be always reported as 0,
although there were errors when deleting the directories/files.

Executing the "rem" command on error works around this
issue. For details see https://stackoverflow.com/a/11137825

Also, remove user prompts and exit only the current script
(exit /b) instead of the whole cmd process to let users call
this script from other scripts for automation purposes for
example.

Signed-off-by: Jens Schmer <jens.schmer@garmin.com>

### DIFF
--- a/build_sdk_windows_qt5_MSVC.bat
+++ b/build_sdk_windows_qt5_MSVC.bat
@@ -75,7 +75,7 @@ echo ************************************
 echo ***  Delete old build Directory  ***
 echo ************************************
 
-    rmdir /s /q build
+    rmdir /s /q build || rem
     if %ERRORLEVEL% NEQ 0 GOTO ERROR_HANDLER
 
 )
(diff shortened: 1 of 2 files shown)


alexmucde:
Hello Jens,

you are right, if you call the script from another script it will stop for user interaction. But that is exactly what i wanted to achieve, so that the windows will not close when started from Windows Explorer and i can see the build results. Perhaps we should great another script, which waits for user input, which calls the current script.
Can you fix this?

     Alex

jschmer-garmin:
Yes, I can add a script called "build_sdk_windows_qt5_MSVC_interactive.bat" that calls "build_sdk_windows_qt5_MSVC.bat" and additionally waits for user input via "set /p name= Continue". Would that be ok?

alexmucde:
Yes this would be great.

jschmer-garmin:
@alexmucde Should be available now. Let me know if that works for you.

PR #253, opened 2022-02-02
Add file explorer feature

petrucianno:
Explore dlt related files from Project view
image

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -6667,26 +6667,51 @@ void MainWindow::on_exploreView_customContextMenuRequested(QPoint pos)
     QMenu menu(ui->exploreView);
     QAction *action;
     /* Get path from index */
-    auto index   = ui->exploreView->selectionModel()->selectedIndexes()[0];
-    auto path    = getPathFromExplorerViewIndexModel(index);
-    bool is_file = !QDir(path).exists();
+    auto indexes   = ui->exploreView->selectionModel()->selectedIndexes();
+    auto index     = indexes[0];
+    auto path      = getPathFromExplorerViewIndexModel(index);
+    bool is_file   = !QDir(path).exists();
 
     if (is_file)
     {
-        action = new QAction("&Load", this);
-         connect(action, &QAction::triggered, this, [this](){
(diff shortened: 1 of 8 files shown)


alexmucde:
Thanks for the contribution, nice improvement.
Build was succesful.
We will test the new feature.

alexmucde (approved):

alexmucde:
I tested your new feature and it looks nice. I will merge soon.
An additional feature would be nice, that you can also select and open several DLT files at once.

alexmucde:
@rus0000 Please also review, then i will merge.

petrucianno:
Hello, @alexmucde.
Additional feature requested was pushed into [file-explorer-feature](https://github.com/petrucianno/dlt-viewer/tree/file-explorer-feature) branch, commit 1346980be94e7e51d31de25994c32bbc5eb76fa7

src/mainwindow.cpp
```diff
@@ -6667,26 +6667,51 @@ void MainWindow::on_exploreView_customContextMenuRequested(QPoint pos)
...
+            QSet<QString> pathsSet;
+            auto selectedIndexes = indexes;
+
+            for (auto &index : selectedIndexes)
+            {
+               QString path = getPathFromExplorerViewIndexModel(index);
+
+               if (path.toLower().endsWith(".dlt"))
+               {
+                   pathsSet.insert(path);
```
alexmucde: Loading multiple file is important to open them in the right order, so that they are sorted by time. QSet does not garantue the order. Please use another class e.g. QList where the order is guaranteed.

alexmucde (changes requested):

src/mainwindow.cpp
petrugrd: Done.
Added possibility to change sorting type and order in explorer view, and the most important - replacing QSet with QList to maintain the right order.

alexmucde (approved):

alexmucde:
Many thanks for the contribution. Sorry for the delay, because i was very busy in the last weeks. I will integrate now and will let you know, if there are still some issues after extensive usage.

PR #273, opened 2022-05-28
Fixed message len for generated message from QDltMsg

AndreySakun:
Fixed message len for generated message from QDltMsg
in case of using sessionId, timestamp and ecuId

Signed-off-by: Andrey Sakun <sakun.andrey@gmail.com>


alexmucde:
Many thanks for the fix!

PR #278, opened 2022-07-05
Enable to set a custom execution sequence for DLT Plugins

gobater:
Currently, the execution sequence of DLT Plugins depends on the discovery/loading order.

According to QDir documentation, all plugins in a certain plugin folder are discovered/loaded alphabetically, but DLTViewer supports more than one plugin folder.

Some Plugins might require some others to be executed before (e.g. in most/some cases it could make sense that Decoder plugins are executed before Viewer Plugins)

This pull request allows the user to set a custom execution sequence for the DLT Plugins and store it in the project file. The plugins can be reordered by:
* context menu
* drag and drop
* dedicated buttons

General/DLT Viewer plugin execution sequence can also be modified in the settings file. However, new execution sequence is not stored automatically when closing the application.

### DIFF
--- a/qdlt/qdltpluginmanager.cpp
+++ b/qdlt/qdltpluginmanager.cpp
@@ -193,6 +193,21 @@ QDltPlugin* QDltPluginManager::findPlugin(QString &name)
     return plugin;
 }
 
+void QDltPluginManager::initPluginPriority(const QStringList& desiredPrio)
+{
+    QStringList finalPrio;
+
+    if(plugins.size() > 1) {
+        int prio = 0;
+        for (int i = 0; i < desiredPrio.count(); ++i) {
+            QString pluginName(desiredPrio[i]);
+            if (setPluginPriority(pluginName, prio)) {
+                ++prio;
+            }
+        }
+    }
+}
+
 bool QDltPluginManager::decreasePluginPriority(const QString &name)
 {
     bool result = false;
@@ -261,6 +276,22 @@ bool QDltPluginManager::setPluginPriority(const QString name, int prio)
     }
 
     return result;
+}
+
+QStringList QDltPluginManager::getPluginPriorities() const
+{
+    QStringList finalPrio;
+
+    if(plugins.size() > 0) {
+        pMutex_pluginList->lock();
+        for(int num=0; num < plugins.size(); num++)
+        {
+            finalPrio << plugins[num]->getName();
+        }
+        pMutex_pluginList->unlock();
+    }
+
+    return finalPrio;
 }
 
 QList<QDltPlugin*> QDltPluginManager::getDecoderPlugins()
@@ -353,3 +384,4 @@ bool QDltPluginManager::initConnections(QStringList list)
 
     return true;
 }
+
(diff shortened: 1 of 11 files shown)


gobater:
@alexmucde , do we need anything else to get this PR approved?

qdlt/qdltpluginmanager.cpp
```diff
@@ -134,108 +151,240 @@ QStringList QDltPluginManager::loadPluginsPath(QDir &dir)
...
         if(plugin->getName()==pluginName)
             plugin->setFilename(filename);
     }
+    pMutex_pluginList->unlock();
 }
 
 
 void QDltPluginManager::decodeMsg(QDltMsg &msg, int triggeredByUser)
 {
+    pMutex_pluginList->lock();
```
mbehr1, line 167: puh. this in the critical path to decode a full file, or? What's the performance impact?

qdlt/qdltpluginmanager.cpp
```diff
@@ -134,108 +151,240 @@ QStringList QDltPluginManager::loadPluginsPath(QDir &dir)
...
-        QDltPlugin *plugin = plugins[num];
+        if(plugins[num]->getName()==name)
+        {
+            plugin = plugins[num];
+            break;
+        }
+    }
+    pMutex_pluginList->unlock();
+
+    return plugin;
```
mbehr1, line 193: this is unsafe, or? does the mutex only protect the order of elements? or can items as well be changed/modified/deleted?

qdlt/qdltpluginmanager.cpp
gobater: good catch, I missed it. I will rework all locks

qdlt/qdltpluginmanager.cpp
gobater: the mutex only protects the order. The elements still can be changed while decoding (as before)

Usha-Rajanna:
MicrosoftTeams-image (5)

Usha-Rajanna:
if we click on first occurrence in the sequence and click down button, this scenario is not working even the right click Move down is not working. 
from second sequence Move up and Down is working.

Usha-Rajanna:
Tested and given the review

qdlt/qdltpluginmanager.cpp
```diff
@@ -13,6 +17,12 @@
 
 QDltPluginManager::QDltPluginManager()
 {
+    pMutex_pluginList = new QMutex();
```
mbehr1, line 20: why as pointer and not as regular member?

qdlt/qdltpluginmanager.cpp
mbehr1, line 167: this still slows down the overall performance of processing every single log, or? Why has this been merged?
What's the performance impact on opening/viewing a bigger file?
@gobater  @Usha-Rajanna

gobater:
@Usha-Rajanna , the remarks from @mbehr1 , were completely valid and I was working to fix them. Why this PR has been merged? This needs to be reverted immediately to avoid performance penalties with large files

alexmucde:
@Usha-Rajanna @Manik-ltts Please perform a performance comparison of version before and after this change, by e.g loading a 1GB dlt file and measure the time. I have not the feeling that the performance has changed.

Usha-Rajanna:
@alexmucde Sure we will check it.

PR #286, opened 2022-08-30
Fix: Release pipeline

agravgaard:
A few minor changes to the MacOS job, so it succeeds, and the artefacts can be uploaded.
It now works on my fork (when a tag is pushed): https://github.com/agravgaard/dlt-viewer/actions/runs/2956350864.

Putting the assets on one line may not be necessary; I can drop that commit if you prefer.

### DIFF
--- a/.github/workflows/Release.yml
+++ b/.github/workflows/Release.yml
@@ -1,3 +1,4 @@
+---
 name: Release
 
 on:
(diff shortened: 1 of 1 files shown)


alexmucde:
Thanks for the contribution. We have no experience with MacOS build. But we will integrate, if we see no problems.

Issue #292, opened 2022-10-21
Fix dead-lock in the dltpluginmanager


svlad-90:
The following change has introduced a regression: https://github.com/COVESA/dlt-viewer/commit/1371752f44e1de25dbad09f6844e680df19c4a21#diff-dc558987f684b77dfc70ddd66149c9cb7d73ab85649f8efa836d3b7c2ed31530

The pMutex_pluginList was added to the QDltPluginManager class in order to sync its calls, which might be done from different threads.

That added mutex is causing a deadlock for some of the plugins.

Imagine the following situation:

- Observed target ECU was rebooted
- Implementation of the dlt-viewer identified that and called 'stateChanged' for a plugin. At that point the 'pMutex_pluginList' is locked once, and control is passed to the plugin. Related file - https://github.com/COVESA/dlt-viewer/blob/master/qdlt/qdltpluginmanager.cpp.

  ```
  bool QDltPluginManager::stateChanged(int index, QDltConnection::QDltConnectionState connectionState,QString hostname)
  {
      pMutex_pluginList->lock(); **// mutex is locked here**
      for(int num=0;num<plugins.size();num++)
      {
          QDltPlugin *plugin = plugins[num];
          if(plugin->isControl() )
              plugin->stateChanged(index,connectionState,hostname); **// control passed to the plugin**
      }
      pMutex_pluginList->unlock();
  
      return true;
  }
  ```

- Inside the plugin's state change, a plugin may call the 'decodeMsg' on a 'QDltMessageDecoder', which was passed to it with the following call from the https://github.com/COVESA/dlt-viewer/blob/master/qdlt/plugininterface.h#L350:
  ```
  // QDltPluginControlInterface
  
  //! A message decoder is injected into the plugin
  /*!
  This function is called by the the viewer once on creation of the plugin
  Motivation - plugin might need to be able to decode a messages, if it directly uses the QDltFile.
  Errors should be reported by providing an error message.
  \sa QDLTPluginInterface::error()
  \param messageDecoder - pointer to a message decoder, which allows a plugin to decode instances of QDltMsg.
  */
  virtual void initMessageDecoder(QDltMessageDecoder* /*messageDecoder*/) = 0;
  ```

- In its turn, QDltPluginManager implements QDltMessageDecoder interface. 'decodeMsg' pure virtual function is implemented in the following way:

  ```
  void QDltPluginManager::decodeMsg(QDltMsg &msg, int triggeredByUser)
  {
[...]

PR #293, opened 2022-10-21
[ISSUE #292] Fix dead-lock in the dltpluginmanager

svlad-90:
- Changed the 'pMutex_pluginList' to work in the recursive mode

Signed-off-by: Vladyslav Goncharuk <svlad1990@gmail.com>


alexmucde:
Thanks for the fix. The solution sounds reasonable. Please integrate quickly.

PR #298, opened 2022-11-09
Filter in selected Index Range

alexmucde:


Manik-ltts:
Unit testing is complete and functionality working as expected.. 
Following are some Observation 
Filter Range Option and start and end index range values are not stored for future use.
Filter Range should be enabled or disable on 'Filters Enabled' check and uncheck.

alexmucde:
This is the idea to reset the index range, whenever a new file is loaded, must not be persistent. Set it to 0 - 0, better would be from 0 - last message number.
Filter Range could be disabled when filters are disabled.

alexmucde:
I will integrate now. Further additions can be added in separate Pull Request,

PR #333, opened 2023-01-19
support mail id updated

Manik-ltts:
Signed-off-by: Manik Fulpagar [manik.fulpagar@ltts.com](mailto:manik.fulpagar@ltts.com)


alexmucde (commented):
This should not be in the OSS version. Please close the PR.

PR #336, opened 2023-01-25
Fix QString from QByteArray length error with Qt6

alexmucde:


Elastino:
@alexmucde Thank you very much. 
Confirmed APID less than 4 characters are properly filtered. 
Tested with length of 2, 3 and 4. (Didn't have a chance to test single though.)

alexmucde:
@Elastino Also many thanks to you for finding the issue, as we are currently starting to support Qt6.
I will merge now.

Issue #345, opened 2023-02-15
DLT Viewer crashes when right click on Explorer tab in some specific empty area


Available12346:

Available12346:
@petrucianno please take a look, thank you

PR #346, opened 2023-02-19
Fix crash when right click on Explorer tab in some specific empty area (#345)

petrucianno:
This commit fixes issue #345


alexmucde:
@petrucianno Thanks for the fix

PR #369, opened 2023-08-12
Build on Apple Silicon

hannesa2:


hannesa2:
Ahh, you merged it !
The build was not successful and doesn't work. 
In my point of view, this is a mistake

PR #373, opened 2023-08-19
No viable conversion from 'QDltDataView' to 'QByteArray'

SangTruongTan:
This approach is working on macOS arm64 architecture.


hannesa2 (approved):

hannesa2:
Sorry, I've no qt experience, and I'm just interested.
What's the benefit of conversion to `QByteArray`, or which issue is solved with it.

SangTruongTan:
It looks like there is a defect with clang on macOS arm64. It could not be converted directly from `QDltDataView` to `QbyteArray` in one line.

PR #390, opened 2023-11-08
Allow passing multiple dlt files on commandline

tohammer:
Treat all arguments with dlt ending that are not used up by another option as log files. Sort multiple files by name to guarantee consistent behavior.

\+ Add missing help text for passing dlt file(s) only


alexmucde:
@tohammer Is it possible to add a signed off statement from you to the commit message?

PR #392, opened 2023-11-22
filetransferplugin: Present save popup once

mfukar:
The plugin's save command created a window for every successful or erroneous attempt to save individual files selected. This is tedious for more than a few files.

Instead, collect errors and present a single pop-up window.

In the future, an alternative might be to configure a selection between this behaviour and a brief message in the status bar.

Fixes a bug where doFLDA was erroneously called on messages ending a transfer ("FLFI..."), causing completed status for a file to be incorrect and display inconsistently.

Adds FiletransferPlugin::doFLFI which emits a new signal so that Form::finishfile_slot can record completion properly and auto-save if needed. updatefile_slot now tracks received packages and completion is ONLY recorded if a FLFI message is present.


alexmucde:
@mfukar Oh sorry i have overseen, you have already done.

PR #402, opened 2023-12-19
Set working directory from cli

tohammer:
Simplifies working from commandline when required files are in current directory.


alexmucde:
@tohammer I will merge now you commit and do the requested changes by my own, because i need further modifications.

PR #403, opened 2023-12-19
Allow mixing project and dlt file arguments directly

tohammer:
Decide which one it is based on extension. Simplifies usage from command line when dlt and project files are grouped in directories.

### DIFF
--- a/qdlt/qdltoptmanager.cpp
+++ b/qdlt/qdltoptmanager.cpp
@@ -342,6 +342,19 @@ void QDltOptManager::parse(QStringList *opt)
             log = true;
             qDebug()<< "Loading logfile " << logFile;
         }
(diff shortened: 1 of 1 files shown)


alexmucde:
@tohammer Please can you also here update the option documentation?
I think the options -l and -p are not needed anymore, DLT files (incl. multiple files) and project files can be now loaded without options.

tohammer:
@alexmucde Should i remove the -l and -p options completely or keep them in code for compatibility? (and ofc will update release info/docsaccordingly)

alexmucde:
@tohammer You can remove them completely

alexmucde:
@tohammer I will merge now you commit and do the requested changes by my own, because i need further modifications.

PR #476, opened 2024-05-29
Add tooltips to logs

qjcina:
When logs are really long it's not possible to see all the contents. The easiest solution is to add simple tooltip to every log.


alexmucde:
Thanks for your contribution.
@qjcina Can you please add a signed off statement to your commit message with your E-Mail Address? See other commits as an example.

qjcina:
@alexmucde Done.

alexmucde:
@qjcina I have found an issue with your PR. Looks like not the correct message is shown in th tooltip. I will provide a fix here #481.

PR #405, opened 2024-01-09
Test CI for QT6

hannesa2:

PR #487, opened 2024-06-13
MacOS qt5 arm64 cmake solution

bunty95:
updated qt for arm64

Still need to verify if "qmake" will work for arm64


hannesa2:
@bunty95 
Thanks a lot !

Btw, there is [qt6](https://github.com/COVESA/dlt-viewer/pull/405) pull request as well

bunty95:
Not a problem @hannesa2 

I will take a look and see if i can work some magic.

PR #509, opened 2024-07-26
Replace search progress dialog with an in-place progress bar

vifactor:
From UX point of view, search progress bar dialog detached from main window is one of the most annoying things for me in the application. With this PR, I replace the hide the input dropdown and replace it a progress bar (+ cancel button) while search is in progress as demonstrated below:

image


alexmucde:
@vifactor There is already a progress bar at the right bottom. Could you use this one instead of a new one?
It is used for load DLT files.
Abort button must be also added there.
For Importing and Exporting it is also used now, but not updated yet, because these functions are still running in main UI Thread. Need to implement new threads for Import and Export.

vifactor:
If I remember correctly (no access to the code ATM), the progress bar at the right bottom is used to display indexing progress. The indexing is run in a separate thread, while search is run in the main thread, so they (can) happen concurrently/in parallel. If it is the case, I'm not sure we can combine both progress reports in a single bar.

alexmucde:
A signal/slot concept is used to separate the access to the progress bar. What we must achieve that no of the parallel running threads can be started together. So only indexing, search, import or export can be started at once. I will try to find a solution if i find some time.

vifactor:
@alexmucde 

> What we must achieve that no of the parallel running threads can be started together.

sorry, I do not understand that...   If computational tasks are "paralellizable", everybody is trying to do the opposite: run them in parallel, in order to utilize all available resources. 

Let us perform the following steps:
1. import relatively big dlt-file
2. immediately start a search 

Currently, we will observe that search progress is reported in a detached dialog, while **in the mean time** in the bottom right corner, the indexing progress is reported. Search triggered by a user and dlt indexing are independent (hence trivially parallelizable) tasks, whose progress is reported in two different progress bars. This my PR does not change anything in that regard, except that instead of reporting search progress into a detached dialog, it reports progress into a specially crafted progress bar as shown in the screenshot below: 

image

> A signal/slot concept is used to separate the access to the progress bar.

sure, signal/slots with Qt::AutoConnection + event loop usually do the right things if signals emitted in one thread and slots are executed in another one, but this is not a concern here. As I mentioned above, IMO reporting progress of the two operations into two different bars is the right thing to do here.

alexmucde:
My idea was to have only one operation running at once, and so using only one common progres bar for all. Concurrency is nice, but the SW is currently not fully designed for this. Sometimes it will crash at the moment.

vifactor:
Yeah, but changing that behavior (running indexing in the main thread) is
whole new story...

ср, 14 серп. 2024 р., 16:05 користувач Alexander Wenzel <
***@***.***> пише:

alexmucde:
Ok i will test your implementation. Integration into a single progress bar can be also done later.

alexmucde:
@vifactor There is a height issue with the Toolbar, it is too high: 
image
Can you fix this?

vifactor:
> There is a height issue with the Toolbar, it is too high

interesting, I did not observe such thing on my machine. I'll try to understand why it is so, thanks for testing

alexmucde:
Great, works now, will merge.

Issue #508, opened 2024-07-24
window command line error


EdagrHW:
![image](https://github.com/user-attachments/assets/8a73c4a8-bf3e-4eee-bb86-ec9450b2dcac)

EdagrHW:
@jeremiah

jeremiah:
Hi @jeremiah - I no longer work for GENIVI (which I guess is called 'COVESA' now.) But I imagine the maintainer of DLT ought to be able to help you here. It looks like a configuration issue to me.

PR #510, opened 2024-07-26
install qoffscreen plugin on windows

vifactor:
https://github.com/COVESA/dlt-viewer/issues/508

offscreen plugin is required since this change: https://github.com/COVESA/dlt-viewer/commit/990abade4cfc19d6c7500fb78673128b1377cc51

Probably using `windeployqt`-tool for packaging could have avoid this bug


vifactor:
I do not have window at the moment so cannot check the change for real.

alexmucde:
@vifactor It is then also missing in build_sdk_windows_qt5_MSVC.bat, please fix also there

vifactor:
@alexmucde since I do not know what exactly that .bat-file does, please have a look, I made the requested change

alexmucde:
@vifactor The script is used to build DLT Viewer  for Windows

PR #519, opened 2024-08-08
fix signal for socket error

vifactor:
The QAbstractSocket::error signal existed in Qt<5.15, but was renamed to QAbstractSocket::errorOccured probably to avoid
https://stackoverflow.com/a/48250710

Since the codebase still uses the old style of connections the compiler could not detect that signal does not exist anymore and error was reported only on runtime:
"qt.core.qobject.connect: QObject::connect:
No such signal QTcpSocket::error(QAbstractSocket::SocketError)"

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -3532,7 +3532,7 @@ void MainWindow::connectECU(EcuItem* ecuitem,bool force)
                 disconnect(ecuitem->socket,0,0,0);
                 connect(ecuitem->socket,SIGNAL(connected()),this,SLOT(connected()));
                 connect(ecuitem->socket,SIGNAL(disconnected()),this,SLOT(disconnected()));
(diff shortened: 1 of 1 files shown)


alexmucde:
Need to be tested before integration. Do you have tested this?

vifactor:
Yes, this is tested with qt5.15 and qt6. It will not work with qt less or
equal to 5.12. if it is important to keep backward compatibility with those
old qt version, this can be done with defines.

ср, 14 серп. 2024 р., 16:08 користувач Alexander Wenzel <
***@***.***> пише:

alexmucde:
Causes now build issues with Ubuntu 20.04, still uses Qt 5.12.8. Still a lot of users using Ubuntu 20.04 on our side,  maybe revert and new PR for later integration?

vifactor:
Ok, I see, I'll did not know 5.12 is still in use. I'll prepare a PR handling properly different Qt versions.

PR #526, opened 2024-09-11
Abi in MacOS artifacts

hannesa2:
It's about to have a x86 and an arm64 version for MacOS
image

image


hannesa2:
@bunty95 are our MacOS build artifact a multi architecture artifacts, which makes this pull request obsolete ?

bunty95:
@hannesa2 current builds are processer specific. 
I don't see how this is different from what we already have available.

hannesa2:
current situation 
image

bunty95:
Yes, we will need this then.

PR #531, opened 2024-09-12
Make code backward compatible with Qt 5.12

vifactor:
Names of the signals changed between Qt 5.12 and 5.15, while we still need to maintain compatibility with 5.12


vifactor:
@alexmucde this should address https://github.com/COVESA/dlt-viewer/pull/519#issuecomment-2345557739

I cannot check this against Qt 5.12, cause it would require to setup the environment, but change is rather trivial.

hannesa2:
you could test it in the pipeline by switching global to 5.12....
image

alexmucde:
Builds in Ubunut 20.04, will merge.

PR #537, opened 2024-09-16
Removed creating tar from build

bunty95:
Reduce the size of macOS artifacts that are created during the build CI/CD pipeline.

Changes:
removed tar command from build as after notarization, binaries are tar'ed

### DIFF
--- a/scripts/darwin/build.sh
+++ b/scripts/darwin/build.sh
@@ -68,4 +68,4 @@ echo "FULL_VERSION=${FULL_VERSION}"
 
 mkdir -p dist
 cp ../scripts/darwin/install.md dist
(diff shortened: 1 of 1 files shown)


hannesa2:
@bunty95 
Just for the records, the containing `DLTViewer.dmg` is broken

bunty95:
I know that the .dmg that is created via cPack is not signed. Need to add an additional step I believe in order to sign .dmg.

could you share error log of the .dmg?

hannesa2:
Sorry to say this.
I don't know how to collect error log, in app "Console" I didn't found anything.
Extracting app from dmg and start it, shows me this
image

Maybe you download it by your own, and have a look on your side ?

PR #543, opened 2024-09-23
Use QCommandLineParser for command line args parsing

vifactor:
As a side effect this PR removes some code duplication which had to be kept in sync + cmd parsing functionality is now handled by the Qt and the logic for combining those args is tested (at least those combinations that are given in examples)

### DIFF
--- a/qdlt/qdltoptmanager.cpp
+++ b/qdlt/qdltoptmanager.cpp
@@ -23,6 +23,7 @@
 
 #include "qdltoptmanager.h"
 #include "version.h"
+
 #include <QDebug>
 #include <QFileInfo>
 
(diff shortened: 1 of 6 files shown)


vifactor:
@alexmucde you asked to add "ready"-label when MR is ready, since I do not have rights to add labels, I notify you with the message, that this MR is ready for review

alexmucde (approved):

alexmucde:
@vifactor I had to reverse this change, as it caused several issues on my side and other users. E.g. the order of the parameters are now important. On my side export to CSV was not working anymore, perhaps also an issue of the order. Please provide a new PR when things are fixed.

alexmucde:
@vifactor Please provide the change also for the tool dlt-commander, if possible.

alexmucde:
@vifactor The export works, but the order was important for the -t option;
 This worls: dlt-viewer input.dlt filter.dlf -t -c out.dlt -d -s, but -t at the end does not work

vifactor:
@alexmucde first of all sorry for the troubles. Thanks for providing the exact failing command, I was about to ask what exactly fails. I'll try to investigate, should not be hard with the unit tests.

alexmucde:
@vifactor I gave you also the "Write" role in the project. Please try if you can now set labels in the PRs and perhaps you can also push branches directly to the project.

vifactor:
I tried: I can now assign labels. Thank you, I appreciate the trust!

vifactor:
@alexmucde I can't reproduce the issue you mentioned. The commands `dlt-viewer input.dlt filter.dlf -t -c out.dlt -d -s` and `dlt-viewer input.dlt filter.dlf -c out.dlt -d -s -t` work exactly the same way on my machine. I use Ubuntu 22.04 with Qt 6.2.4

Could you please specify your OS, Qt version and how exactly the command misbehaves for you. If possible, also provide several exact commands that failed.

PR #544, opened 2024-09-26
Fix qmake build

alexmucde:


alexmucde:
@vifactor #528 was not building anymore with qmake, this fixies the change

PR #549, opened 2024-09-29
MacOS release with codesign and notarization

hannesa2:
This is missing in release


hannesa2:
In test run https://github.com/COVESA/dlt-viewer/actions/runs/11093376306 it works now

PR #551, opened 2024-09-29
Minor tablemodel cleanup

vifactor:
mainly code duplication removal. Much easier to review commit-by-commit. 

This activty revealed two directions for further improvements:
- explicit LRU caching in the model
- refactor searchtable model, which is basically a copy of this tablemodel

### DIFF
--- a/src/tablemodel.cpp
+++ b/src/tablemodel.cpp
@@ -144,7 +144,6 @@ TableModel::TableModel(const QString & /*data*/, QObject *parent)
               }
          }
 
-         QString visu_data;
          switch(index.column())
          {
          case FieldNames::Index:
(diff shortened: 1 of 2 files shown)


src/tablemodel.cpp
```diff
@@ -27,69 +27,70 @@
...
 
-     long int filterposindex = 0;
-
+     auto decodeMessageWithPlugin = [&] {
+         if (QDltSettingsManager::getInstance()->value("startup/pluginsEnabled", true).toBool())
+         {
+             if ( decodeflag == 1 )
+             {
+                 decodeflag = 0;
+                 pluginManager->decodeMsg(msg, !QDltOptManager::getInstance()->issilentMode());
```
vifactor, line 85: @alexmucde initially I had a look to this file to understand if I can remove usage of QDltOptManager singleton in here. this is the only place it is used. Do you know how issilentmode command option relates to triggeredByUser argument of the decodeMsg? Can I simply pass `false` here? I know it is not used in the existing plugins, however I am not sure about third-party ones

PR #565, opened 2024-10-14
Fix and Improve RegEx Replace feature.

alexmucde:
RegEx Replace will only be executed when filter matches. Re enable RegEx Replace for DLT Export, no influence on non verbose messages anymore.


else2623:
image
image

PR #567, opened 2024-10-14
Add dlf file extension when saving filter

vifactor:
If user-typed filename does not have any extension.

This small UX improvement I decided to propose after spending some time to understand why "default filter functionality" does not work even though I placed the filter file into right folder. The cause was that I saved filter in a file without dlf-suffix

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -6795,16 +6795,24 @@ void MainWindow::filterAdd()
 
 void MainWindow::on_action_menuFilter_Save_As_triggered()
 {
-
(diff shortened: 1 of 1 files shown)


vifactor:
@alexmucde how would you want me to use the rights you granted to me? May I merge small PRs like this one which I am confident about?

alexmucde:
@vifactor I would like to have a final look before merging, or even test the version if it is a bigger thing. But my time is limited, so i can have only a look from time to time.

alexmucde (approved):

vifactor:
makes sense, of course, I just wanted to clarify

PR #569, opened 2024-10-21
Simplify handling of logging only mode in log table view

vifactor:
With this MR logging only mode is handled separately before any message decoding happens. Before this MR, although msg was default-assigned in one of the if-cases, it still was non-empty due to multiple decodings happening in the data-method for values of `role`-variable other than `DisplayRole`.

### DIFF
--- a/src/tablemodel.cpp
+++ b/src/tablemodel.cpp
@@ -28,8 +28,6 @@
 #include "qdltoptmanager.h"
 
 static long int lastrow = -1; // necessary because object tablemodel can not be changed, so no member variable can be used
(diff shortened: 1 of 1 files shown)


alexmucde:
@vifactor Are you able to test this in an Online Logging use case?

alexmucde:
I have tried it with enabling and disabling this feature and it works, at least in the offline mode.

vifactor:
I tried in offline too. I will check it in online a bit later

vifactor:
online mode seem to work too:
image

PR #574, opened 2024-10-25
Fix for command line parser issue with multiple arguments with same t…

alexmucde:
…ype.


alexmucde:
@vifactor Now find a small issue, but was easily to be fixed. Using the options -b and -e mutiple times was not possible anymore. Changing the used function from value to values solved the issue.

vifactor:
Great! Thanks

PR #580, opened 2024-11-17
Fix wheel scrolling of the main tableview

vifactor:
The introduction of the Ctrl+Wheel combination broke the normal mouse table view scrolling. This fixes the issue


alexmucde:
@vifactor Thanks for the quick fix.

PR #586, opened 2024-11-20
Fix compilation issues

omircon:
# Commit 1: Add missing includes

**Why:**
Without the missing includes, we had incomplete types.

**What:**
Added the necessary includes to resolve the incomplete types issue.

# Commit 2: Set C standard to c11

**Why:**
`dlt_common.c` includes `<QtGlobal>` which implicitly includes `<qtypes.h>`, which contains a `static_assert` that requires the C11 standard. This was causing issues with the GCC compiler, which does not support `static_assert` in earlier C standards.

**What:**
Explicitly set the C standard to C11 to ensure compatibility with the `static_assert` usage in the code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,8 @@ set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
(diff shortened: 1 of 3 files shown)


LocutusOfBorg:
thanks for fixing, Ubuntu build was just bitten by this issue
https://launchpadlibrarian.net/763274632/buildlog_ubuntu-plucky-amd64.dlt-viewer_2.27.0+dfsg-1_BUILDING.txt.gz

Issue #576, opened 2024-11-12
UI is unresponsive fo seconds at the end of dlt load


vifactor:
Long running MainWindow::reloadLogFileFinishFilter() is executed in main thread and hangs UI

vifactor:
This https://github.com/COVESA/dlt-viewer/blob/3f51d3f71145c5a5407a969f9cfebe620275abca/src/mainwindow.cpp#L2068 helps to avoid UI freeze, but unfortunately it is not thread safe as far as I can see and currently emits a lot of warnings.

The real fix as I can see it will include significant refactoring of the `MainWindow::contextLoadingFile` function.

vifactor:
Refactored code a bit here.

Apparently, parsing messages takes much less than populating the tree, i.e. constructing and updating UI elements
getLogInfoMessages took  196 ms
populateEcusTree took  4309 ms

vifactor:
@alexmucde currently when I import 2 dlt files sequentially, in the logs table only the logs from the second file become seen. I would expect that in such case the ECUs tree from the "Config"-tab should correspond to the latest imported file. This doesn't however happen, the ECUs tree gets appended as far as I can see, so it contains ECUs data from both files. Is this bug or feature?

alexmucde:
@vifactor Issue solved and integrated, can be closed?

vifactor:
@alexmucde issue is not resolved, the PR you merged is a little improvement. The real fix is more complicated, I have a draft for it, but needs cleanup. Please keep this issue open

If possible, could you give me the answer for the question above: https://github.com/COVESA/dlt-viewer/issues/576#issuecomment-2508651277 ?

vifactor:
@alexmucde the PR is in place, it fixes the issue. I am not sure, though I tested it well for bad inputs, since I do not have a collection of test cases with various inputs. Maybe we can merge it, people start using it and when/if there is a regression found we fix it? I do not see any other way to validate that not-so trivial PR

alexmucde:
Following pull request can solve the issue, please try:
https://github.com/COVESA/dlt-viewer/pull/633

Issue #203, opened 2021-04-23
Build AppId/CtxId for all received DLt messages, not only Register messages


alexmucde:

alexmucde:
Any contributer available to work on this?

vifactor:
I could have a look, but I need more details. Could you give me a link to the code where "only build from register Context DLT Ctrl message" happens?

PR #606, opened 2024-12-23
Do not block UI when building ECUs tree

vifactor:
This PR resolves https://github.com/COVESA/dlt-viewer/issues/576 for the cases when DLT log contains lots of apps and contexts. 

Historically, the same function used to populate ECU tree (in the "Config"-tab) both in online and offline modes. This was a problem for offline mode (aka dlt-file loading), because UI update is then called multiple time in a short time frame ate the end of file processing freezing the whole app. The fix is to:
1. Separate parsing of control messages payloads from UI creation
2. When tree view is created use more efficient constructors, which do not force many redundant UI updates.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -84,6 +84,7 @@ extern "C" {
 #include "sortfilterproxymodel.h"
 #include "qdltoptmanager.h"
 #include "qdltctrlmsg.h"
+#include "ecutree.h"
 
 
 MainWindow::MainWindow(QWidget *parent) :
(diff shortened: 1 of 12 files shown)


src/mainwindow.cpp
```diff
@@ -2066,12 +2075,21 @@ void MainWindow::reloadLogFileFinishFilter()
...
+                auto ctrlMsg = qdlt::msg::payload::parse(msg.getPayload(), msg.getEndianness() == QDltMsg::DltEndiannessBigEndian);
+                std::visit([&ecuTree, ecuId = msg.getEcuid()](auto&& payload) {
+                    using T = std::decay_t<decltype(payload)>;
+                    if constexpr (std::is_same_v<T, qdlt::msg::payload::GetLogInfo>) {
+                        ecuTree.add(ecuId, payload);
+                    }
+                }, ctrlMsg);
+            }
         }
+        project.ecu->clear();
```
vifactor, line 2082: @alexmucde this changes this behavior: https://github.com/COVESA/dlt-viewer/issues/576#issuecomment-2508651277 I.e. now every new dlt file import will clean up the ECU tree first and create new one only with its own entries.

This makes this change a bit simpler, unless old behavior was intentional.

vifactor:
@alexmucde I would appreciate if you could play with this. I tried not to bring any regressions, but as you can see the change is not small, so I might have introduced some. Also I would appreciate if you could provide data for tests to make them more robust, e.g. raw data to check parsing of other types of payloads except `GetLogInfo`

alexmucde:
@vifactor It would be good, if we can also solve this issue with this PR:
https://github.com/COVESA/dlt-viewer/issues/203
The idea is that not only Control messages are used to build the ECU Tree, also App/CtxIds of all messages are used to build the tree.
We have currently an ECU, where no Control messages are used at all.

vifactor:
@alexmucde ok, I'll have a look, I now better understand that issue. Keep in mind, it might take some time, while this PR already fixes one very real annoying issue.

PR #608, opened 2025-01-04
Fix export rules header

vifactor:
The first commit is the most important, it fixes includes that I missed here: https://github.com/COVESA/dlt-viewer/pull/594

Remove a bit of dead code in follow up commits. Can be reviewed commit-by-commit


alexmucde (approved):

alexmucde:
@vifactor I have integrated the plugin interface changes, but know i have a lot of issues in our internal plugins, similar to what you have changed already in the OSS plugins. Now i have two options. Revert your changes or walk through each issue and fix it.

vifactor:
@alexmucde you mean you need to do this type of changes in your private plugins, right?
image

If so, give me a couple of hours to think how this can be fixed with minimal nb of changes, because I find changes made in this MR are good. I understand the pain of fixing multiple places, so if you think it is easier, feel free to revert of course, and sorry for that, I did not expect I would break API

alexmucde:
I see you changes are useful, but will break the interface. I will first give it a try to fix it in our plugins. If it works easily we can keep your changes. I hope a find and replace will help.

vifactor:
okay, so the easiest way to fix your plugins IMO and not revert these changes is to simply define:
```
class QDltMsg {
  static QString toAsciiTable(...) {
      return QDlt::toAsciiTable(...);
  }
}
```

I think this should do the job

alexmucde:
The first things were easily fixed. But you also have removed some array functions, which i have used in a plugin:
C:\workspace\github\covesa\dlt-viewer_plugin_development\plugin\adnonverboseplugin\adnonverboseplugin.cpp:1110: Fehler: Use of undeclared identifier 'qDltMessageType'
C:\workspace\github\covesa\dlt-viewer_plugin_development\plugin\adnonverboseplugin\adnonverboseplugin.cpp:1115: Fehler: Use of undeclared identifier 'qDltLogInfo'; did you mean 'QDltMsg::DltLogInfo'? (fix available)
C:\workspace\github\covesa\dlt-viewer_plugin_development\plugin\adnonverboseplugin\adnonverboseplugin.cpp:1115: Fehler: Subscripted value is not an array, pointer, or vector
C:\workspace\github\covesa\dlt-viewer_plugin_development\plugin\adnonverboseplugin\adnonverboseplugin.cpp:1118: Fehler: Use of undeclared identifier 'qDltTraceType'
C:\workspace\github\covesa\dlt-viewer_plugin_development\plugin\adnonverboseplugin\adnonverboseplugin.cpp:1121: Fehler: Use of undeclared identifier 'qDltNwTraceType'
C:\workspace\github\covesa\dlt-viewer_plugin_development\plugin\adnonverboseplugin\adnonverboseplugin.cpp:1124: Fehler: Use of undeclared identifier 'qDltControlType'; did you mean 'QDltMsg::DltControlTime'? (fix available)
C:\workspace\github\covesa\dlt-viewer_plugin_development\plugin\adnonverboseplugin\adnonverboseplugin.cpp:1124: Fehler: Subscripted value is not an array, pointer, or vector
C:\workspace\github\covesa\dlt-viewer_plugin_development\plugin\adnonverboseplugin\adnonverboseplugin.cpp:1183: Fehler: Use of undeclared identifier 'qDltTypeInfo'
C:\workspace\github\covesa\dlt-viewer_plugin_development\plugin\adnonverboseplugin\adnonverboseplugin.cpp:1190: Fehler: Use of undeclared identifier 'qDltTypeInfo'

vifactor:
oh, I could not imagine those vars were part of the API. They moved to cpp files, fix should be also easy: just copy them back to QDlt-header, but keep `inline` keyword

alexmucde:
I have fixed it now by adding the vars to the plugin as you have done.

The following line still makes some issues qdltbase.h:
inline constexpr const auto DLT_MAX_MESSAGE_LEN = 1024*64;

When i remove "inline" it works, looks like it is incompatible with older C++ standards and our plugins are build with older C++ standard. I will create a PR for this.

PR #611, opened 2025-01-07
Minor plugins refactoring

vifactor:
Mainly removal unnecessary heap allocations (use stack instead), redundant code removal, use of algos instead of raw loops

### DIFF
--- a/qdlt/qdltplugin.h
+++ b/qdlt/qdltplugin.h
@@ -71,7 +71,7 @@ public:
     bool isCommand();
 
     // generic plugin interfaces
-    QString name();
+    QString name() const;
     QString pluginVersion();
     QString pluginInterfaceVersion();
     QString error();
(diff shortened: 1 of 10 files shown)


qdlt/qdltpluginmanager.cpp
```diff
@@ -131,60 +118,39 @@ QStringList QDltPluginManager::loadPluginsPath(QDir &dir)
...
+    std::for_each(plugins.begin(), plugins.end(), [&](auto* plugin) {
+        if (plugin->name() == pluginName)
             plugin->setFilename(filename);
-    }
-    pMutex_pluginList->unlock();
+    });
 }
 
-
 void QDltPluginManager::decodeMsg(QDltMsg &msg, int triggeredByUser)
```
vifactor, line 129: this is probably the only function which is kinda used in multithreaded context, it is hence the only one that has to be protected by the mutex.

qdlt/qdltplugin.h
```diff
@@ -71,7 +71,7 @@ class QDLT_EXPORT QDltPlugin
     bool isCommand();
 
     // generic plugin interfaces
-    QString name();
+    QString name() const;
```
vifactor, line 74: many of the methods here should actually be `const`, but I do not change that because I do not know what promise we have to hold for plugins implementers

alexmucde (approved):

PR #613, opened 2025-01-10
Update qdltbase.h

alexmucde:
Make it compatible again with C++ standard below C++17.


vifactor:
looks good, at the end `constexpr` is implicitly `inline`

Issue #616, opened 2025-01-21
Search matches if payload of message is empty


alexmucde:
Currently also DLT messages with empty payload will be shown in the search results.

vifactor:
is this something I can help with? does it show messages with empty payload when searching empty messages, i.e. query string is empty? 

would be nice to have a test dlt file if possible

alexmucde:
Yes  you can help here. I have created a Tool for another purpose, but this generates some DLT messages, also empty ones with some modifications. I have added this file. If you now search for Relais3, you get also the empty messages in the search results.

[20250121_140442_DLTRelaisPluginTest_Logs.zip](https://github.com/user-attachments/files/18491280/20250121_140442_DLTRelaisPluginTest_Logs.zip)

vifactor:
So far cannot reproduce:

![Image](https://github.com/user-attachments/assets/b183fb55-11d0-44c1-ac38-08d8723abc2a)

Tried on Linux v.2.27.0 and current master head. Will try Windows later. Maybe you have some non-default settings?

alexmucde:
Hmm my output looks like this, including my Search settings in the Screenshot:

![Image](https://github.com/user-attachments/assets/0464d586-e904-4a62-a6e0-642c709b520b)

alexmucde:
When Regular Expressions is active, the empty lines are gone.

vifactor:
thanks, indeed with regular expressions off could reproduce. Will try to find time to understand why this happens

vifactor:
@alexmucde PR with fix is up

PR #618, opened 2025-01-23
Fix incorrect matching algo

vifactor:
https://github.com/COVESA/dlt-viewer/issues/616

It was probably a copy-paste issue


alexmucde:
@vifactor Thanks very much for the quick solution.

Issue #363, opened 2023-07-13
Commandline and Export function for splitting files


alexmucde:
An option should be added to split DLT files during export and by commandline.

hannesa2:
> ... and by commandline.

Maybe `split` is the command you are looking for.
As I don't know what you want to do _exactly_, I've only an advice https://www.networkworld.com/article/3489256/breaking-linux-files-into-pieces-with-the-split-command.html

vifactor:
I am working on this, but only in scope of commander

vifactor:
PR has been merged and now commander has an option to split dlt files.

PR #621, opened 2025-02-02
Dlt files split

vifactor:
https://github.com/COVESA/dlt-viewer/issues/363

### DIFF
--- a/commander/CMakeLists.txt
+++ b/commander/CMakeLists.txt
@@ -14,7 +14,9 @@
 
 add_executable(dlt-commander
     main.cpp
-    optmanager.cpp)
+    optmanager.cpp
+    dltfileexporter.h
+    dltfileexporter.cpp)
 
 target_link_libraries(dlt-commander
     qdlt
--- a/commander/commander.pro
+++ b/commander/commander.pro
@@ -30,8 +30,9 @@ CONFIG(debug, debug|release) {
 
 SOURCES += \
         main.cpp \
-        optmanager.cpp
+        optmanager.cpp \
+        dltfileexporter.cpp
 
 HEADERS += \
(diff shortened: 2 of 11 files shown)


alexmucde:
Interesting feature. Can you also add the begin/end Data/Time to the filename of the split files?
Would be good if we can split the file already when importing e.g. from MF4.

vifactor:
> Can you also add the begin/end Data/Time to the filename of the split files?

sure, in what format you want to have that?

> Would be good if we can split the file already when importing e.g. from MF4.

I can do that in a separate PR, but I would need from you an example of MF4 file. I never use those and have no samples.

alexmucde:
> > Can you also add the begin/end Data/Time to the filename of the split files?
> 
> sure, in what format you want to have that?
> 
We have something like this in our logger files: 2024-10-22_22-24-47_2024-10-22_22-26-47_
It will be best to add this to the beginning of the file for better sorting.

> > Would be good if we can split the file already when importing e.g. from MF4.
> 
> I can do that in a separate PR, but I would need from you an example of MF4 file. I never use those and have no samples.
Bad i a cannot share our MF4 files.

vifactor:
> Bad i a cannot share our MF4 files.

I have found some random mf4 files in Internet. However, I do not understand how to process them with the commander. This line https://github.com/COVESA/dlt-viewer/blob/c3b63978cf01a902327cc819937bce92911cc917/commander/main.cpp#L41 prevents me from importing only mf4, a dlt file must be importer alongside:
```
ERROR: No DLT file used. At least one DLT file must be provided.
```

It also command line example given for `--help` option does not work: `dlt-commander -c output.txt input1.mf4 input2.mf4` is not actually working, right?

PR #630, opened 2025-02-17
Copy Message Payload

Pavithra-aa-Anand:
Shortcut implementation for Copy message payload

Signed-off by : Pavithra Anand <Pavithra.AA.Anand@partner.bmwgroup.com>

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -271,10 +271,15 @@ MainWindow::~MainWindow()
     delete m_shortcut_searchnext;
     delete m_shortcut_searchprev;
     delete sortProxyModel;
+    delete copyPayloadShortcut;
 }
 
 void MainWindow::initState()
 {
+    /* Shortcut for Copy Selection Payload to Clipboard */
+    copyPayloadShortcut = new QShortcut(QKeySequence("Ctrl+P"), this);
+    connect(copyPayloadShortcut, SIGNAL(activated()), this, SLOT(onActionAenuConfigCopyPayloadToClipboardTriggered()));
+
     /* Settings */
     settingsDlg = new SettingsDialog(&qfile,this);
     settingsDlg->assertSettingsVersion();
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -279,7 +279,7 @@ void MainWindow::initState()
 {
     /* Shortcut for Copy Selection Payload to Clipboard */
     copyPayloadShortcut = new QShortcut(QKeySequence("Ctrl+P"), this);
-    connect(copyPayloadShortcut, &QShortcut::activated, this, &MainWindow::onActionAenuConfigCopyPayloadToClipboardTriggered);
+    connect(copyPayloadShortcut, &QShortcut::activated, this, &MainWindow::onActionMenuConfigCopyPayloadToClipboardTriggered);
 
     /* Settings */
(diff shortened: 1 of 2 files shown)


src/mainwindow.cpp
```diff
@@ -271,10 +271,15 @@ MainWindow::~MainWindow()
     delete m_shortcut_searchnext;
     delete m_shortcut_searchprev;
     delete sortProxyModel;
+    delete copyPayloadShortcut;
```
vifactor: you pass `this` as parent when creating the `copyPayloadShortcut` instance which is the Qt way to do memory management, so this `delete` leads to double free issue most probably

src/mainwindow.cpp
```diff
@@ -271,10 +271,15 @@ MainWindow::~MainWindow()
...
 }
 
 void MainWindow::initState()
 {
+    /* Shortcut for Copy Selection Payload to Clipboard */
+    copyPayloadShortcut = new QShortcut(QKeySequence("Ctrl+P"), this);
+    connect(copyPayloadShortcut, SIGNAL(activated()), this, SLOT(onActionAenuConfigCopyPayloadToClipboardTriggered()));
```
vifactor: would be nicer to use modern version of `connect` instead of old macro-based. The former is checked by compiler

src/mainwindow.cpp
Pavithra-aa-Anand: Understood. will remove this line

src/mainwindow.cpp
Pavithra-aa-Anand: Will rewrite this line using modern version of connect syntax.

src/mainwindow.cpp
```diff
@@ -275,6 +275,10 @@ MainWindow::~MainWindow()
 
 void MainWindow::initState()
 {
+    /* Shortcut for Copy Selection Payload to Clipboard */
+    copyPayloadShortcut = new QShortcut(QKeySequence("Ctrl+P"), this);
+    connect(copyPayloadShortcut, &QShortcut::activated, this, &MainWindow::onActionAenuConfigCopyPayloadToClipboardTriggered);
```
vifactor:
```suggestion
    connect(copyPayloadShortcut, &QShortcut::activated, this, &MainWindow::onActionMenuConfigCopyPayloadToClipboardTriggered);
```

is this a misprint in ui file somewhere?

vifactor:
The shortcut works, however I suggest to add hints to the context menu and menu, so that users could easily learn about this new shortcut. What I mean that context menu shown on the mouse right-click contains `Copy selection on clipboard` and File menu contains "Copy to clipboard" entries, those should now additionally mention `Ctrl + P`

src/mainwindow.cpp
Pavithra-aa-Anand: No. I didn't make any change in the name of the slot. This slot name was already there (onAction**Aenu**ConfigCopyPayloadToClipboardTriggered instead of onAction**Menu**ConfigCopyPayloadToClipboardTriggered).

vifactor: would it be possible for you to fix this misprint?

Pavithra-aa-Anand: Ok sure. Will fix it.

vifactor (approved):
looks good, please squash your commits

PR #635, opened 2025-02-18
Persist Search History

Pavithra-aa-Anand:
Persist the search history even across DLT Viewer restarts.

Signed-off by : Pavithra Anand <Pavithra.AA.Anand@partner.bmwgroup.com>

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -50,6 +50,7 @@
 #include <QtEndian>
 #include <QDir>
 #include <QDirIterator>
+#include <QCompleter>
 
 /**
  * From QDlt.
@@ -100,6 +101,8 @@ MainWindow::MainWindow(QWidget *parent) :
     setAcceptDrops(true);
 
     target_version_string = "";
+
+    loadSearchHistoryList(searchHistory);
 
     initState();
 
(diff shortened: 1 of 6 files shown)


src/mainwindow.cpp
```diff
@@ -8292,6 +8305,19 @@ void MainWindow::onSearchProgressChanged(bool isInProgress)
     ui->dockWidgetProject->setEnabled(!isInProgress);
 }
 
+void MainWindow::loadSearchHistoryList(QStringList& searchHistory)
```
vifactor: Could you please move this whole stuff into the corresponding widget (search form or search dialog) instead of extending main window which is already a polluted superclass antipattern?

vifactor (changes requested):

src/mainwindow.cpp
Pavithra-aa-Anand: Ok, will move it to search dialog

src/mainwindow.h
```diff
@@ -166,10 +166,12 @@ class MainWindow : public QMainWindow
     QStringList autoloadPluginsVersionStrings;
 
     /* String List Containing Search History */
-    enum { MaxSearchHistory = 50 };
+    enum { MaxSearchHistory = 20 };
     QAction *searchHistoryActs[MaxSearchHistory];
     QStringList searchHistory;
-
+    QLineEdit* searchLineEdit;
```
vifactor: can all this be also moved to the corresponding widget?

Pavithra-aa-Anand: Yes, sure. Will move to search form/search dialog

Pavithra-aa-Anand: I have already utilised searchHistory which is QStringList in saving and loading history functions. 
In searchForm, I could see m_historyModel which is QStringListModel is defined. 

So, Is it expected to proceed with searchHistory or m_historyModel?

Pavithra-aa-Anand: Is it fine to proceed with (QStringList) searchHistory.

vifactor: I cannot tell without seeing a code. Take an approach, provide a code and I'll review it

PR #637, opened 2025-02-19
Configure Keyboard Shortcuts

Pavithra-aa-Anand:
Configured all the keyboard shortcuts of DLT Viewer

Signed-off by : RenuPriya Krishnamoorthy <renupriya.ka.krishnamoorthy@partner.bmwgroup.com>


src/mainwindow.cpp
```diff
@@ -5625,6 +5625,78 @@ void MainWindow::on_action_menuHelp_Command_Line_triggered() {
...
+
+    // Fill the model with data (unchanged)
+    QStringList names = {"New", "Open", "Save As", "Clear", "Import DLT Stream",
+                         "Import DLT Stream with serial header", "Find", "Jump To",
+                         "New Project", "Open Project", "Save Project", "Expand All ECU",
+                         "Collapse All ECU", "Copy Payload", "Info", "Quit"};
+    QStringList shortcuts = {"Ctrl + N", "Ctrl + O", "Ctrl + S", "Ctrl + E", "Ctrl + I",
+                             "Ctrl + J", "Ctrl + F", "Ctrl + G", "Ctrl + Shift + G",
+                             "Ctrl + Shift + O", "Ctrl + Shift + S", "Ctrl++", "Ctrl+",
+                             "Ctrl + P", "F1", "Ctrl + Q"};
```
vifactor: what about defining this in a list of pairs to automatically ensure that counts match and have a nicer loop below?

src/mainwindow.cpp
```diff
@@ -5625,6 +5625,78 @@ void MainWindow::on_action_menuHelp_Command_Line_triggered() {
                 QDltOptManager::getInstance()->getHelpText());
 }
 
+void MainWindow::on_actionShortcuts_List_triggered(){
+    qDebug() <<"Shortcuts Triggered";
+
+    QDialog *shortcutDialog = new QDialog(this);
```
vifactor, line 5523: you are going to create new dialog every time this function is triggered, show it once and keep it on the heap whole app lifecycle with no possibility to show it again, effectively leaking the memory

vifactor (changes requested):

src/mainwindow.cpp
Renu-Priya411: Instead of two different lists, you are asking me to make a single list of pairs. Is my understanding of your feedback correct ?

vifactor: correct

src/mainwindow.cpp
Renu-Priya411: Creating a static pointer will assure that QDialog is created only once and allocating a memory for it will handle memory leaking as well. May i know if your opinion on this, if i can proceed forward with this idea ?

vifactor: Making it static will work, but then you'll need to wrap whole creation logic into `if (dialog)` to check if it was already created.

My approach (in order of preference) would be:
1. Create the dialog on the stack in this function
2. Have this dialog as the member of mainwindow class, create it once upon mainwindow creation

Renu-Priya411: yes, I didnt wrap the whole logic in if(dialog). I have given a return if the dialog is already opened and the logic continues. I will try to implement your approach as well. Thank u for the input

src/mainwindow.cpp
```diff
@@ -5547,6 +5548,69 @@ void MainWindow::on_action_menuHelp_Command_Line_triggered() {
                 QDltOptManager::getInstance()->getHelpText());
 }
 
+void MainWindow::on_actionShortcuts_List_triggered(){
+    qDebug() <<"Shortcuts Triggered";
+
+    shortcutDialog.setWindowTitle("Shortcuts List");
+    shortcutDialog.resize(600, 400);
+
+    QTableView *table = new QTableView(&shortcutDialog);
```
vifactor: with the way you changed it, you still "leak" memory for these allocations. My suggestion was to define shortcutDialog on the stack, but not as a class member

Renu-Priya411: Okay. In that case, QDialog will be initialised inside the the function as "QDialog shortcutDialog(this)". Using this case dialog will be destroyed once its out of scope. Please let me know if my understanding correct ?

vifactor:
```
void MainWindow::on_actionShortcuts_List_triggered(){
    QDialog shortcutDialog;
    [...]
    shortcutDialog.exec();
}
```

this is what I mean

src/mainwindow.cpp
```diff
@@ -5547,6 +5548,69 @@ void MainWindow::on_action_menuHelp_Command_Line_triggered() {
                 QDltOptManager::getInstance()->getHelpText());
 }
 
+void MainWindow::on_actionShortcuts_List_triggered(){
+    qDebug() <<"Shortcuts Triggered";
```
vifactor:
```suggestion
```

src/mainwindow.cpp
```diff
@@ -5547,6 +5548,69 @@ void MainWindow::on_action_menuHelp_Command_Line_triggered() {
...
+        }
+    }
+    table->setModel(model);
+    table->setColumnWidth(0, 275);
+    table->setColumnWidth(1, 275);
+    QVBoxLayout *layout = new QVBoxLayout(&shortcutDialog);
+    layout->addWidget(table);
+
+    shortcutDialog.setLayout(layout);
+    shortcutDialog.show();
```
vifactor:
```suggestion
    shortcutDialog.exec();
```

src/mainwindow.cpp
```diff
@@ -5547,6 +5548,69 @@ void MainWindow::on_action_menuHelp_Command_Line_triggered() {
...
 
+void MainWindow::on_actionShortcuts_List_triggered(){
+    qDebug() <<"Shortcuts Triggered";
+
+    shortcutDialog.setWindowTitle("Shortcuts List");
+    shortcutDialog.resize(600, 400);
+
+    QTableView *table = new QTableView(&shortcutDialog);
+    table->setObjectName("Shortcuts Summarise Table");
+    QStandardItemModel *model = new QStandardItemModel(0, 2, this);
```
vifactor:
```suggestion
    QStandardItemModel *model = new QStandardItemModel(0, 2, &shortcutDialog);
```

src/mainwindow.cpp
```diff
@@ -5547,6 +5548,69 @@ void MainWindow::on_action_menuHelp_Command_Line_triggered() {
...
+
+    QFont BoldFont;
+    BoldFont.setBold(true);
+
+    QStandardItem *headerName = new QStandardItem("Name");
```
vifactor:
```suggestion
    QStandardItem *headerName = new QStandardItem("Action name");
```

src/mainwindow.cpp
```diff
@@ -5547,6 +5548,69 @@ void MainWindow::on_action_menuHelp_Command_Line_triggered() {
...
+
+    shortcutDialog.setWindowTitle("Shortcuts List");
+    shortcutDialog.resize(600, 400);
+
+    QTableView *table = new QTableView(&shortcutDialog);
+    table->setObjectName("Shortcuts Summarise Table");
+    QStandardItemModel *model = new QStandardItemModel(0, 2, this);
+
+    QFont BoldFont;
+    BoldFont.setBold(true);
```
vifactor:
```suggestion
    QFont headerFont;
    headerFont.setBold(true);
```

src/mainwindow.cpp
```diff
@@ -5547,6 +5548,69 @@ void MainWindow::on_action_menuHelp_Command_Line_triggered() {
...
+    QStandardItem *headerName = new QStandardItem("Name");
+    headerName->setFont(BoldFont);
+    model->setHorizontalHeaderItem(0, headerName);
+
+    QStandardItem *headerFeature = new QStandardItem("Shortcuts");
+    headerFeature->setFont(BoldFont);
+    model->setHorizontalHeaderItem(1, headerFeature);
+
+    // Define shortcut list using QPair
+    QList<QPair<QString, QString>> shortcutsList = {
```
vifactor, line 5561: how can you ensure that this list of shortcuts remains valid over the time? For example, some contributor can come and change `"Copy Payload", "Ctrl + P"` to `"Copy Payload", "Ctrl + J"`, then list will be outdated, but it would be hard to notice.

Renu-Priya411: Shall I keep the table in an editable format, so that any contributor can edit it if incase there is any change ? Remaining changes I have already made.

vifactor: my point is that you define shortcut combinations in two different unrelated places (where is `Ctrl + P` first mentioned?) with time this will lead to the issue, that this table contains outdated info because somebody did not fix it when changing shortcut combination. I do not see how making the table editable can help here. What might help is to define a central place where all combinations are defined like variables and use the corresponding variable in every place where it appears

Renu-Priya411: for E.g : 
QString Payload = "Ctrl + P"; 
QString New = " Ctrl + N";

The variables will be defined in a separate place and Payload, New, etc  will be used for creating a summarised table. If any changes are made to the existing shortcuts this variable has to changed within the code and if a new shortcut is created a new variable has to be added. Is this the expected way of implementation ?

vifactor: something like that, yes

Renu-Priya411: const QString shortcutNew = "Ctrl + N";

    const QString shortcutOpen = "Ctrl + O";

    const QString shortcutSave = "Ctrl + S";

    const QString shortcutClear = "Ctrl + E";

    // Store shortcuts dynamically using a list of pairs

    QList<QPair<QString, QString>> shortcutsList = {

        {"New", shortcutNew},

        {"Open", shortcutOpen},

        {"Save As", shortcutSave},

        {"Clear", shortcutClear}

    };
    
    This is how the implementation looks. If any changes are made the variable has to be changed and then it has to be added to the QList. This process is more or less similar to the old one. In the old method it was one QList and here we have variables which are added to the QList. To display all the shortcuts in a table
, I want all the data in one container. Hence all the shortcuts has to be present in a consolidated containers like QList or QMap. May i know is it okay to go with this implementation ?

vifactor: please provide a commit with the changes, it is hard to see what you are going to do exactly from the comment

src/mainwindow.cpp
```diff
@@ -5551,6 +5552,98 @@ void MainWindow::on_action_menuHelp_Command_Line_triggered() {
...
+    const QString shortcutimportDLT = "Ctrl + I";
+    const QString shortcutimportSerialHeader = "Ctrl + J";
+    const QString shortcutFind = "Ctrl + F";
+    const QString shortcutJumpTo = "Ctrl + G";
+    const QString shortcutNewProject = "Ctrl + Shift + G";
+    const QString shortcutOpenProject = "Ctrl + Shift + O";
+    const QString shortcutSaveProject = "Ctrl + Shift + S";
+    const QString shortcutExpandAllECU = "Ctrl++";
+    const QString shortcutCollapseAllECU = "Ctrl+";
+    const QString shortcutCopyPayload = "Ctrl + P";
```
vifactor, line 5556: having this list here does not solve the problem I outlined, because for example this combination of keys exists also here: https://github.com/COVESA/dlt-viewer/blob/3394aad838a235134d2ea99ed8bc3c67ff128f6e/src/mainwindow.cpp#L281 and can be changed independently leading to inconsistencies between this new dialog and actual shortcut.

Renu-Priya411: Hello @vifactor 
I have added the list here because, I have to put all the shortcuts to a container and then utilise that container for displaying the shortcuts which are present in the code. Can you please elaborate more on what is the expected changes ?

alexmucde:
In the top left of the Shortcuts Window there are some scrollbars visible, whichg should not be there, please fix:
image

Issue #638, opened 2025-03-09
string arguments with ascii encoding are wrongly decoded as utf8 strings


mbehr1:
String arguments with ascii encoding (!(DLT_TYPE_INFO_SCOD == DLT_SCOD_UTF8)),  DltTypeInfoStrg) are decoded as utf8 strings (DltTypeInfoUtf8).

The typeInfo is correctly determined here:
```
if (dltType& DLT_TYPE_INFO_STRG)
    {
        if ((dltType & DLT_TYPE_INFO_SCOD)==DLT_SCOD_UTF8)
        {
            typeInfo = DltTypeInfoUtf8;
        }
        else
        {
            typeInfo = DltTypeInfoStrg;
        }
    }
```
but in
https://github.com/COVESA/dlt-viewer/blob/b7577fbb7fe503fd93dab7fd749bc6a2557c7ded/qdlt/qdltargument.cpp#L394

the QString::QString(QByteArray&) constructor is used that (since Qt5) uses fromUtf8.
Should be changed to use QString::fromLatin1(...).

I attached a test dlt file that shows the problem.
It contains two messages. Each with two arguments: a string with utf8 explaining what's expected and then the ascii string.

It should parse as
```
0 1970/01/01 01:00:00.000000          0 000 ECU1 adlt txta log fatal verbose 2 next arg is an ascii string with 'übe': übe
1 1970/01/01 01:00:00.000000          0 000 ECU1 adlt txta log fatal verbose 2 next arg is an ascii string with '€÷¢µ©': €÷¢µ©
```

but dlt-viewer 2.27.0 (macOS, arm) parses it as:
```
0 1970/01/01 01:00:00.000000 0.0000 0 ECU1 adlt txta 0 log  verbose 2 next arg is an ascii string with 'übe': �be
1 1970/01/01 01:00:00.000000 0.0000 0 ECU1 adlt txta 0 log  verbose 2 next arg is an ascii string with '€÷¢µ©': �����
```

mbehr1:
[test_ascii_utf8_strings.zip](https://github.com/user-attachments/files/19151485/test_ascii_utf8_strings.zip)

mbehr1:
(I guess I could create better examples that even show where the ascii sequences is a valid utf8 char (so completely wrong content then)

mbehr1:
With PR#639 under windows/arm it looks like:
 
[...]

PR #639, opened 2025-03-10
fix: use fromLatin1 for string args with ascii encoding

mbehr1:
The default QString constructor uses internally fromUtf8 which is wrong for strings that claim to be ascii encoded.

Issue: #638

### DIFF
--- a/qdlt/qdltargument.cpp
+++ b/qdlt/qdltargument.cpp
@@ -391,12 +391,12 @@ QString QDltArgument::toString(bool binary) const
         break;
     case DltTypeInfoStrg:
         if(data.size()) {
-            text += QString::fromLatin1(data.data());
+            text += QString::fromLatin1(data);
         }
         break;
     case DltTypeInfoUtf8:
         if(data.size()) {
-            text += QString::fromUtf8(data.data());
+            text += QString::fromUtf8(data);
         }
         break;
     case DltTypeInfoBool:
@@ -533,12 +533,12 @@ QVariant QDltArgument::getValue() const
         break;
     case DltTypeInfoStrg:
         if(data.size()) {
-            return QVariant(QString(getData()));
+            return QVariant(QString::fromLatin1(data));
         }
         break;
     case DltTypeInfoUtf8:
         if(data.size()) {
-            return QVariant(QString::fromUtf8(data.data()));
+            return QVariant(QString::fromUtf8(data));
         }
         break;
     case DltTypeInfoBool:
@@ -646,7 +646,7 @@ bool QDltArgument::setValue(QVariant value, bool verboseMode)
         return true;
     case QVariant::String:
         data = value.toByteArray();
-        typeInfo = QDltArgument::DltTypeInfoStrg;
+        typeInfo = QDltArgument::DltTypeInfoUtf8;
         return true;
     case QVariant::Bool:
         {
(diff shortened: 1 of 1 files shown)


mbehr1:
@alexmucde can you review and approve? thx!

mbehr1:
With this PR under windows/arm the test file from #638 looks like:
0 1970/01/01 01:00:00.000000 0.0000 0 ECU1 adlt txta 0 log  verbose 2 next arg is an ascii string with 'übe': übe
1 1970/01/01 01:00:00.000000 0.0000 0 ECU1 adlt txta 0 log  verbose 2 next arg is an ascii string with '€÷¢µ©': ÷¢µ©

(need to check which font is used so that the ascii/latin1 € sign is not shown.)

qdlt/qdltargument.cpp
```diff
@@ -391,7 +391,7 @@ QString QDltArgument::toString(bool binary) const
         break;
     case DltTypeInfoStrg:
         if(data.size()) {
-            text += QString("%1").arg(QString(getData()));
+            text += QString::fromLatin1(data.data());
```
vifactor:
```suggestion
            text += QString::fromLatin1(data);
```

should work the same way

vifactor (approved):
I am not sure what encoding `DltTypeInfoStrg` assumes, but apparently for your case it works. The way it is currently handled in master, is anyway weird.

qdlt/qdltargument.cpp
mbehr1: hmm. I copied it from the utf8 path a few lines below that uses the same .data() method that returns a char* (or const char*).
Checking the qt doc this makes a difference:
a) .data() always ensures it's zero terminated. (so it will add a \0 if needed). Might not be needed here as QString::fromLatin1 could handle it.
b) QString::fromLatin1(const char *str, int size = -1) (if size is -1 then it takes strlen(...). which stops at the first 0. So if a 0 is in the middle it will behave differently.

So not the same. But thinking about it I do prefer if zeros are not stripped.
And Qt6.2 seems to change the signature of the function to be: QString::fromLatin1(const char *str, qsizetype size) (so no default param for size).

-> I will change it to use only (data) (will do the same for the fromUtf8 as well.
@vifactor thx for the review!

mbehr1:
@vifactor  see update. tested as well on win/arm. Looks similar to first commit.
(I noticed that the dltviewerplugin ascii tab doesn't show the ascii data (neither the utf8). Will take a look at that separately. It's not a regression of the first/2nd commit.)

vifactor:
thanks for your contribution, would it be maybe possible also to add some unit tests for the functionality?

vifactor:
also please add `Signed-off-by: ...` entry into your commit messages

mbehr1:
@vifactor  sure. I can add a few unit tests.
For the "Signed-off-by: ..." is it ok for upcoming commits or shall I add to the prev. 2 as well?
Is a single commit preferred or 2-3 different ones?

vifactor:
> For the "Signed-off-by: ..." is it ok for upcoming commits or shall I add to the prev. 2 as well?

Not anymore necessary for this PR, but normally, Alex requests it for each commit.

> I can add a few unit tests.

still welcome in next PR

mbehr1:
added unit tests in #641

PR #641, opened 2025-03-15
tests: add unit test for QDltArgument

mbehr1:
Add first unit tests for QDltArgument class focussing on utf8/ascii strings.

### DIFF
--- a/qdlt/tests/test_qdltargument.cpp
+++ b/qdlt/tests/test_qdltargument.cpp
@@ -1,0 +1,71 @@
+#include <gtest/gtest.h>
+
+#include <qdltargument.h>
+
+TEST(QDltArgument, constructor) {
+    QDltArgument arg;
+
+    ASSERT_EQ(arg.getTypeInfo(), QDltArgument::DltTypeInfoUnknown);
+    ASSERT_EQ(arg.getTypeInfoString(), QString{});
+    ASSERT_EQ(arg.getDataSize(), 0);
+    ASSERT_EQ(arg.toString(), QString("?")); // a bit weird for data size 0
+    ASSERT_EQ(arg.getValue(), QVariant());
+}
+
+TEST(QDltArgument, string_ascii) {
+    QDltArgument arg;
+    arg.setTypeInfo(QDltArgument::DltTypeInfoStrg);
+    const unsigned char asc_data[] = {
+        0xc3, 0xbc, 'b', 'e' // Ã¼be (and not übe)
+    };
+    arg.setData(QByteArray::fromRawData((const char*)asc_data, sizeof(asc_data)));
+    ASSERT_EQ(arg.toString(), QString::fromUtf8("Ã¼be"));
+
+    // parse from raw payload
+    const unsigned char asc_data2[] = {0x00, 0x02, 0x00, 0x00, // type info
+                                       0x04, 0x00,             // str len
+                                       0xc3, 0xbc, 'b',  'e'};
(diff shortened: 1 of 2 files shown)


mbehr1:
@vifactor see here the unit tests for the QDltArgument utf8/ascii change.

qdlt/tests/test_qdltargument.cpp
```diff
@@ -0,0 +1,71 @@
...
+    const unsigned char asc_data2[] = {0x00, 0x02, 0x00, 0x00, // type info
+                                       0x04, 0x00,             // str len
+                                       0xc3, 0xbc, 'b',  'e'};
+    QByteArray payload = QByteArray::fromRawData((const char*)asc_data2, sizeof(asc_data2));
+    unsigned int offset = 0;
+    arg.setArgument(payload, offset, QDlt::DltEndiannessLittleEndian);
+    ASSERT_EQ(arg.getTypeInfo(), QDltArgument::DltTypeInfoStrg);
+    ASSERT_EQ(arg.getDataSize(), 4);
+    ASSERT_EQ(arg.toString().length(), 4);
+    ASSERT_EQ(arg.toString(), QString::fromUtf8("Ã¼be"));
```
vifactor, line 34: I think second implies the first

qdlt/tests/test_qdltargument.cpp
```diff
@@ -0,0 +1,71 @@
...
+    QDltArgument arg;
+    arg.setTypeInfo(QDltArgument::DltTypeInfoUtf8);
+    const unsigned char asc_data[] = {
+        0xc3, 0xbc, 'b', 'e' // übe
+    };
+    arg.setData(QByteArray::fromRawData((const char*)asc_data, sizeof(asc_data)));
+    ASSERT_EQ(arg.toString(), QString::fromUtf8((const char*)asc_data, sizeof(asc_data)));
+
+    // parse from raw payload
+    const unsigned char asc_data2[] = {0x00, 0x82, 0x00, 0x00, // type info
```
vifactor, line 49: maybe create QByteArray directly? this way you will save on casting from raw C-arrays

vifactor (approved):
thanks for adding tests!

qdlt/tests/test_qdltargument.cpp
mbehr1: yep, but I wanted to make it explicit. in the 2nd test the length for same binary payload is 3.

qdlt/tests/test_qdltargument.cpp
mbehr1: QByteArray has the same (signed) char instead of unsigned char constructor issue, doesn't it?

vifactor: I thought about something like this: 
```
QByteArray payload = QByteArray::fromHex("000200000400c3bc6265");
```

vifactor: @mbehr1 would you mind making it nicer as I suggested?

mbehr1: @vifactor I thought about it but I find it more difficult to read as an indenting and commenting of byte slices would not be so easy.
I could change to a 
```
"...fromHex("000" // foo
 + "1234" // bar
+ "...." // xyz
); 
```
But that's not really nice either?

vifactor: ok, fair enough

PR #645, opened 2025-03-25
Fix deprecated types

pgawro:
- qdltargument.cpp -> switch with deprecated QVariant replaced with QMetaType
 - plugintreewidget.cpp -> replaced event->pos with event->position->toPoint

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,4 @@ release/
 *.orig
 *~
 .idea
-
+.cache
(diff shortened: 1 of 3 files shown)


pgawro:
might help with   #405

vifactor:
@pgawro thanks for your contribution. Unfortunately, the app should still work with Qt5.15 (and probably even 5.12). Those methods that you replaced are just deprecated in Qt6, so either you go with somewhat ugly `#ifdef Qt6` or we live with deprecation warnings until complete transition happens.

vifactor (changes requested):
fix the build for Qt5 or drop changes

pgawro:
@vifactor ahh yes you are right. So for now I think is better to drop those changes.

PR #646, opened 2025-03-30
fix crash due to wrongly parsed ctrl message

vifactor:
GetSoftwareVersion ctrl message does not encode
string length in the payload

Additionally, protect parsing function
with try-catch

See
https://github.com/COVESA/dlt-viewer/discussions/642


alexmucde:
Looks fine to me

PR #650, opened 2025-04-09
prevent memory leak on each table view context menu open

vifactor:
before this change every menu open created multiple QActions on the heap. Because the main window was
assigned as a parent for those, the memory was deallocated at the end of the program while created actions were not usable anymore.


vifactor:
Windows failure is related to https://github.com/jurplel/install-qt-action/issues/281

Issue #667, opened 2025-06-03
Save and Open Project Settings does not restore ECU connections


alexmucde:
In the current v2.28.0 Beta it is not possible to Save and Open again the project settings. The ECU settings are not visible anymore. Seams like a regression in one of the changes since 2.27.0
Saving the project settings in v2.28.0 and Opening with v2.27.0 is possible. Looks like the project file itself is ok.

vifactor:
Ok, I'll have a look

vifactor:
@alexmucde I provided a fix, however I am not sure it does not affect some special workflows.

alexmucde:
Fixed and merged

PR #668, opened 2025-06-09
Fix ECUs config tree when importing dlp

vifactor:
This is a fix of regression introduced with
https://github.com/COVESA/dlt-viewer/commit/a7bd8af1fffc9711e535704183853bedbd8e1c14

The fix prevents running the indexer when importing dlp, otherwise indexer calls reloadLogFileFinishFilter which recreates empty ecu tree

https://github.com/COVESA/dlt-viewer/issues/667


alexmucde (approved):
Tested and issue is fixed

PR #670, opened 2025-06-09
Use Qt6 for macos 14 and 15 CI job

vifactor:


.github/workflows/BuildPR.yml
```diff
@@ -106,6 +106,35 @@ jobs:
         with:
           name: DLTViewer-${{ matrix.macos }}-${{ matrix.abi }}
           path: build/dist/DLTViewer*.tgz
+  
+  buildMacOS15:
```
hannesa2: Sorry, with this way, you have a lot of boiler plate code in github action.

hannesa2 (changes requested):
Please consider to extend matrix. Preparation is here #675
And in my point of view, a move to qt6 should also happen on Linux and Windows too

hannesa2:
Please rebase to recent master and here #677 you have an idea, how to extend the matrix

vifactor:
> And in my point of view, a move to qt6 should also happen on Linux and Windows too

I agree, but not necessarily simultaneously. Btw, migration to Qt6  has already happened on ubuntu 24.04. 

> Btw, there is an other draft too https://github.com/COVESA/dlt-viewer/pull/405

I know, that MR was too scary to touch. My goal was simply add a new job to check macos-15 + Qt6 build without packaging, only because I do not have macos so it would be close to impossible for me to check produced package even if I figure out how to produce one.

vifactor:
@hannesa2 if you have a mac machine, this PR now generates a mac bundle, would you be able to check whether app can be launched from the bundle?

.github/workflows/BuildPR.yml
```diff
@@ -166,3 +166,72 @@ jobs:
             build/dist/DLTViewer*.7z
             build/dist/DLTViewer*.exe
             build/dist/*.md
+
+  buildMacOS15Qt6:
```
hannesa2: Using the matrix and not introduce a new job, would avoid the start problems on MacOS. 
Signing and notarization is in matrix out of the box

bunty95:
If RPATH is toggled to "ON" it should correct some of qt libraries being embedded into the .app file
Possibly we can use this template with QT5 and integrate notarization.

.github/workflows/BuildPR.yml
```diff
@@ -166,3 +166,72 @@ jobs:
...
+    - name: Configure project with CMake
+      run: |
+        mkdir build
+        cd build
+        echo "QT_ROOT_DIR=$QT_ROOT_DIR"
+        APP_DIR_NAME="DLTViewer.app"
+        cmake -G Ninja -DCMAKE_PREFIX_PATH=$QT_ROOT_DIR/lib/cmake \
+        -DCMAKE_OSX_DEPLOYMENT_TARGET=12.0 \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DDLT_USE_QT_RPATH=OFF \
```
bunty95, line 152: DDLT_USE_QT_RPATH=ON

vifactor: I spent quite some time to find out that this option leads to some rpath garbage with qt6 and macos 15, which in turn leads to macdeployqt tool failure

.github/workflows/BuildPR.yml
vifactor: Using matrix is good to save on code, but in this case important thing is to make build with qt6 working. The problem with existing packaging code is that it is crafted for qt5, is full of obsolete and hardcoded assumptions which are hard to fix in such a way that it works both with qt5 amnd qt6. Here, I am trying to come up with a minimal setup which would allow me to produce a working macos 15 bundle with qt6

bunty95:
Added notarization to this PR. It now runs on without issue.

hannesa2:
> Added notarization to this PR. It now runs on without issue.

I see no reason to duplicate this. With an new `build_cmake_qt6.sh` and add it the matrix you probably can avoid this boilerplate code

vifactor:
@bunty95 @hannesa2 I do not have mac, but because I desperately want us to migrate to qt6, I made this work blindly relying on ci. Further progress on my side is not possible. please help me first to fix the bundle, then we can decide how to proceed

bunty95:
Also do we need macOS 13, 14, and 15? I think we can drop macOS 14 since arm64 is captured by macOS 15, and intel is captured by macOS 13.

hannesa2:
For my point of view, when qt6 is working, we don't need qt5 anymore.
But I don't want to decide this

hannesa2:
> Also do we need macOS 13, 14, and 15? I think we can drop macOS 14 since arm64 is captured by macOS 15, and intel is captured by macOS 13.

The I've no opinion.
Important is, that we have arm64 and x86. But even x86 is questionable.

bunty95:
> > Also do we need macOS 13, 14, and 15? I think we can drop macOS 14 since arm64 is captured by macOS 15, and intel is captured by macOS 13.
> 
> There I've no opinion. But currently nothing hurt here. Important is, that we have arm64 and x86. But even x86 is questionable.

Once apple drops x86 from macOS, we can drop it from project

bunty95:
> I tested it on macOS 15.5.
> 
> I can open the app file Screenshot 2025-06-29 at 18 36 15
> 
> About the dmg file, it doesn't work if I just run it (in dmg or move it into `Applications` folder) Screenshot 2025-06-29 at 18 37 23 Because from macOS 15.1, you can't run unsigned applications on macOS. https://www.osnews.com/story/141055/bug-or-intentional-macos-15-1-completely-removes-ability-to-launch-unsigned-applications/ https://hackaday.com/2024/11/01/apple-forces-the-signing-of-applications-in-macos-sequoia-15-1/
> 
> I don't know how this application codesigning works. However, I can test running this app with some tweek. You can run the app in DMG file after run this command: `xattr -c <your path>/DLTViewer.app && codesign --force --deep -s - <your path>/DLTViewer.app` Screenshot 2025-06-29 at 18 49 05

Thank you for double checking. using the "--force" is not the correct way to sign a ".dmg".
I will look into the correct method to do this, but for now the ".app" will be bundled, along with a release on the appstore.

vifactor:
@hannesa2 @bunty95 @nickelpot could somebody try the bundle after @bunty95 added signing and notarization? According to what @nickelpot said, this should now work without issues on macos-15

bunty95:
I think we can merge this into the main branch.
I would recommend removing QT5 from the macOS build complete.

vifactor:
> I think we can merge this into the main branch.

I somewhat share the @hannesa2 's concern here https://github.com/COVESA/dlt-viewer/pull/670#discussion_r2157443083 So I guess I will try to squeeze these changes into that matrix job if you confirm that signed dmg from this branch already works fine.

scripts/darwin/codesign.sh
```diff
@@ -0,0 +1,26 @@
...
+security unlock-keychain -p "$MACOS_CI_KEYCHAIN_PWD" build.keychain
+security import certificate.p12 -k build.keychain -P "$MACOS_CERTIFICATE_PWD" -T /usr/bin/codesign
+security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k "$MACOS_CI_KEYCHAIN_PWD" build.keychain
+
+/usr/bin/codesign --timestamp --options=runtime -s "$MACOS_CERTIFICATE_NAME" -f -v /Users/runner/work/dlt-viewer/dlt-viewer/build/install/DLTViewer.app/Contents/Frameworks/* 
+/usr/bin/codesign --timestamp --options=runtime -s "$MACOS_CERTIFICATE_NAME" -f -v /Users/runner/work/dlt-viewer/dlt-viewer/build/install/DLTViewer.app/Contents/Resources/* 
+/usr/bin/codesign --timestamp --options=runtime -s "$MACOS_CERTIFICATE_NAME" -f -v /Users/runner/work/dlt-viewer/dlt-viewer/build/install/DLTViewer.app/Contents/PlugIns/networkinformation/* 
+/usr/bin/codesign --timestamp --options=runtime -s "$MACOS_CERTIFICATE_NAME" -f -v /Users/runner/work/dlt-viewer/dlt-viewer/build/install/DLTViewer.app/Contents//PlugIns/iconengines/* 
+/usr/bin/codesign --timestamp --options=runtime -s "$MACOS_CERTIFICATE_NAME" -f -v /Users/runner/work/dlt-viewer/dlt-viewer/build/install/DLTViewer.app/Contents/PlugIns/imageformats/*
```
hannesa2: Does it works instead of
`/Users/runner/work/dlt-viewer/...`
also with this ?
`~/work/dlt-viewer/...`

vifactor: if your idea was to remove the full hardcoded path, I did it, but simply using relative path.

I would appreciate if you could try the signed dmg from this new job with Qt6 to check whether it finally works.

.github/workflows/BuildPR.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   buildMac:
-    name: Build macOS
+    name: Build ${{ matrix.macos }} (${{ matrix.abi }}) with Qt${{ matrix.qt }}
```
hannesa2, line 11: When you remove the matrix elements, then the matrix will be added to name by default.
Please see 
image

.github/workflows/BuildPR.yml
```diff
@@ -53,53 +44,16 @@ jobs:
...
-          /usr/bin/codesign --timestamp --options=runtime -s "$MACOS_CERTIFICATE_NAME" -f -v /Users/runner/work/dlt-viewer/dlt-viewer/build/install/DLTViewer.app/Contents/PlugIns/styles/*
-          /usr/bin/codesign --timestamp --options=runtime -s "$MACOS_CERTIFICATE_NAME" -f -v /Users/runner/work/dlt-viewer/dlt-viewer/build/install/DLTViewer.app/Contents/PlugIns/virtualkeyboard/* 
-          /usr/bin/codesign --timestamp --options=runtime -s "$MACOS_CERTIFICATE_NAME" -f -v /Users/runner/work/dlt-viewer/dlt-viewer/build/install/DLTViewer.app/Contents/MacOS/plugins/*
-          /usr/bin/codesign --timestamp --options=runtime -s "$MACOS_CERTIFICATE_NAME" -f -v /Users/runner/work/dlt-viewer/dlt-viewer/build/install/DLTViewer.app/Contents/MacOS/dlt-commander 
-          /usr/bin/codesign --timestamp --options=runtime -s "$MACOS_CERTIFICATE_NAME" -f -v /Users/runner/work/dlt-viewer/dlt-viewer/build/install/DLTViewer.app/Contents/MacOS/dlt-viewer 
-          /usr/bin/codesign --timestamp --options=runtime -s "$MACOS_CERTIFICATE_NAME" -f -v /Users/runner/work/dlt-viewer/dlt-viewer/build/install/DLTViewer.app
+        run: scripts/darwin/codesign.sh
```
hannesa2, line 47: Nice, this is the baseground to reuse codesign and notarize in other apps as well.
But for this, the scripts need more generalization

scripts/darwin/notarize.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+echo "Create keychain profile"
+xcrun notarytool store-credentials "notarytool-profile" --apple-id "$PROD_MACOS_NOTARIZATION_APPLE_ID" --team-id "$PROD_MACOS_NOTARIZATION_TEAM_ID" --password "$PROD_MACOS_NOTARIZATION_PWD"
+echo "Creating temp notarization archive"
+ditto -c -k --keepParent "build/install/DLTViewer.app" "build/install/DLTViewer.zip"
```
hannesa2, line 6: It's not a must, but when 
`"build/install/DLTViewer.app"` `"build/install/DLTViewer.zip"` are parameters, this script can be uses easily in other repos as well

scripts/darwin/codesign.sh
```diff
@@ -0,0 +1,38 @@
...
+security default-keychain -s build.keychain
+security unlock-keychain -p "$MACOS_CI_KEYCHAIN_PWD" build.keychain
+security import certificate.p12 -k build.keychain -P "$MACOS_CERTIFICATE_PWD" -T /usr/bin/codesign
+security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k "$MACOS_CI_KEYCHAIN_PWD" build.keychain
+
+VERSION=$(sw_vers -productVersion)
+MAJOR=$(echo "$VERSION" | cut -d '.' -f 1)
+if [[ "$MAJOR" -ge 14 ]]; then
+  echo "macOS is Sonoma (14) or later"
+  /usr/bin/codesign --timestamp --options=runtime -s "$MACOS_CERTIFICATE_NAME" -f -v build/install/DLTViewer.app/Contents/PlugIns/networkinformation/*
```
hannesa2, line 19: It's not a must, but when 
`"build/install/DLTViewer.app"` is a parameter, this script can be uses easily in other repos as well

hannesa2 (approved):
Yes !
I can start it 
image

This is out of range of this pull request:
What is `unstable` ? This is in my point of view not self explaining 

Much better would be the output of `git describe --tags` here
image

scripts/darwin/notarize.sh
bunty95: the last line "build/install/DLTViewer.app" "build/install/DLTViewer.zip" is a must otherwise Apple will reject the signed app. Apple only accepts ".zip" files for verification.
Ditto is used since it preserves the data once in the zip format.

bunty95:
> > macos 13 x86 remained on qt5
> 
> Does it mean, that for x86 there is no qt5 possible at all ?

x86 can keep qt5, but there is no need since qt6 will still work on x86.
it will simplify the build. It will also allow only 2 builds to occur (arm and intel)
I believe that using the macos 14 runner will build both (macos-latest , macos-14-latest)

scripts/darwin/notarize.sh
hannesa2: Ok, I thought it was clear, that I mean that as suggestion this `build/install/DLTViewer.app` `build/install/DLTViewer.zip` should be parameters from given shell script and should be used as `$1` and `$2`

bunty95: Yes that could work, didn't understand what you meant originally.
As long as the output from the build ".app" is the same as $1 and $2, is the same as $1 just with ".zip" it will work.

PR #673, opened 2025-06-17
Windows 2019 to latest

bunty95:
Change windows 2019 to windows 2025. Will impact both PR and Release workflows.

This is due to Github no longer supporting Windows 2019

### DIFF
--- a/.github/workflows/BuildPR.yml
+++ b/.github/workflows/BuildPR.yml
@@ -139,7 +139,7 @@ jobs:
       matrix:
         task: [ sdk ]
         qt: [ 5 ]
-    runs-on: windows-2019
+    runs-on: windows-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


hannesa2:
Sorry, I can only help in CI issues and I don't use Windows.
But when it builds, (in my point of view) all is fine !
Are there unit tests ?

.github/workflows/BuildPR.yml
```diff
@@ -139,7 +139,7 @@ jobs:
       matrix:
         task: [ sdk ]
         qt: [ 5 ]
-    runs-on: windows-2019
+    runs-on: windows-latest
```
hannesa2, line 142: I learned that `latest` can cause issues in the future.
Maybe suddenly it will stop working.
Please consider to change to `windows-2022`

bunty95: I can change it to windows-2025. 
That way it wont have to change for a couple of years?
I will pull a new PR for this.

alexmucde:
@bunty95 Still using 2019 on my machine, this change breaks my build. Perhaps we can change the script that both wil work.

PR #679, opened 2025-06-20
Windows 2025

bunty95:
change windows-latest to windows-2025, as issues can occur with windows-latest

### DIFF
--- a/.github/workflows/BuildPR.yml
+++ b/.github/workflows/BuildPR.yml
@@ -139,7 +139,7 @@ jobs:
       matrix:
         task: [ sdk ]
         qt: [ 5 ]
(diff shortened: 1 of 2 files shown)


.github/workflows/BuildPR.yml
```diff
@@ -139,7 +139,7 @@ jobs:
       matrix:
         task: [ sdk ]
         qt: [ 5 ]
-    runs-on: windows-latest
+    runs-on: windows-2025
```
hannesa2: Maybe an interim step with 2022 makes sense here ?

bunty95: update made. I will merge the PR

PR #682, opened 2025-06-23
Smoketest on MacOS build

hannesa2:


hannesa2:
This makes no sense on build machine. 
Here the test of signing and notarization makes no sense. It would need an other machine

PR #689, opened 2025-07-10
Prepare Releas 2.28.0

alexmucde:


hannesa2:
I would change the label to something like "Switch back to Windows 2019"

ReleaseNotes_Viewer.txt
```diff
@@ -15,6 +15,94 @@ Known issues
 Changes
 -------
 
+2.28.0
+
+* Configure Keyboard Shortcuts (#637)
```
hannesa2, line 20: Still the same hint: This is obsolete, because release job generates this automatically

alexmucde: I made some cleanup here.

PR #694, opened 2025-07-18
DLT File Splitting Settings

Renu-Priya411:
The given dlt file can be split into smaller sizes and can be saved in local as per user's convenience

### DIFF
--- a/src/filespliting.cpp
+++ b/src/filespliting.cpp
@@ -1,0 +1,203 @@
+#include <mainwindow.h>
+#include <filespliting.h>
+
+#include "qmessagebox.h"
+#include <QDialog>
+#include <QFileDialog>
+#include <qboxlayout.h>
+#include <qlabel.h>
+#include <qlineedit.h>
+#include <qpushbutton.h>
+
+FileSpliting::FileSpliting(QWidget *parent)
+    : QWidget(parent){
+
+}
+
+void FileSpliting::setFile(QFile *file)
+{
+    m_file = file;
+}
+
+
+//The function is triggered when Split DLT File is clicked which is under File Menu.
+//A dialog box asking for the splitting size is opened where the user can specify the size in KB,MB or GB.
+//The size is casted toqint64 and them parsed to splitOutputFile function.
+//The destination path can also be selected by user which will also be parsend to splitOutFile function.
+
+void FileSpliting::splitDLTFile_triggered(QFile &file,QStringList path){
+    qDebug() << "Split File Triggered";
+
+    QDialog dialog(this);
+    dialog.setWindowTitle("Enter Split Size");
+
+    QLabel *sizeLabel = new QLabel("Size (number):");
+    QLineEdit *sizeEdit = new QLineEdit();
+    sizeEdit->setValidator(new QDoubleValidator(0, 999999, 2, this)); // Accepts decimal input
+
+    QLabel *unitLabel = new QLabel("Unit:");
+    QComboBox *unitCombo = new QComboBox();
+    unitCombo->addItem("KB");
+    unitCombo->addItem("MB");
+    unitCombo->addItem("GB");
+
+    QPushButton *okButton = new QPushButton("OK");
+    QPushButton *cancelButton = new QPushButton("Cancel");
+
+    QHBoxLayout *inputLayout = new QHBoxLayout();
+    inputLayout->addWidget(sizeLabel);
+    inputLayout->addWidget(sizeEdit);
+    inputLayout->addWidget(unitLabel);
+    inputLayout->addWidget(unitCombo);
+
+    QHBoxLayout *buttonLayout = new QHBoxLayout();
+    buttonLayout->addStretch();
+    buttonLayout->addWidget(okButton);
+    buttonLayout->addWidget(cancelButton);
+
+    QVBoxLayout *mainLayout = new QVBoxLayout();
+    mainLayout->addLayout(inputLayout);
+    mainLayout->addLayout(buttonLayout);
+    dialog.setLayout(mainLayout);
+
+    connect(okButton, &QPushButton::clicked, &dialog, &QDialog::accept);
+    connect(cancelButton, &QPushButton::clicked, &dialog, &QDialog::reject);
+
+    if (dialog.exec() == QDialog::Accepted) {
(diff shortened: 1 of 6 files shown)


alexmucde:
- Split files only at the end of a DLT message
- Implement not in Mainwindow.cpp, is getting too large

src/filespliting.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+        QFile output(outputFileName);
+        if (output.open(QIODevice::WriteOnly)) {
+            output.write(buffer);
+            output.close();
+            qDebug() << "Written last split:" << QDir::toNativeSeparators(outputFileName);
+        }
+    }
+
+    m_file->close();
```
shubhamshahaBMW, line 244: when trying the feature, getting error "Unexpected EOF: file ends before message fully read." in application output. 

Also, last few logs (20-100 lines) are missed. Data loss is seen when splitting the files.
Please test and correct the behaviour.

shubhamshahaBMW (changes requested):

src/filespliting.cpp
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    int fileIndex = 1;
+    qint64 accumulatedSize = 0;
+    QByteArray buffer;
+
+           // Reset to start
+    m_file->seek(0);
+
+    while (!m_file->atEnd()) {
```
shubhamshahaBMW, line 167: Also this while look blocks UI thread, so if we try to split big file, the whole DLT-Viewer application is stuck and does not respond till the export is done. (which might take ~30-40 sec. in some cases.)
may be having thread safe implementation for this might help. if not, we can create a pop-up which show's the progress while export is going on.

shubhamshahaBMW (changes requested):

src/filespliting.cpp
```diff
@@ -0,0 +1,203 @@
...
+            QFileDialog::ShowDirsOnly | QFileDialog::DontResolveSymlinks
+            );
+
+        if (folderPath.isEmpty()) {
+            QMessageBox::warning(this, "No Folder Selected", "Split operation canceled.");
+            return;
+        }
+
+
+        if (folderPath.isEmpty()) {
```
shubhamshahaBMW: redundant empty checks.

src/filespliting.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+    connect(okButton, &QPushButton::clicked, &dialog, &QDialog::accept);
+    connect(cancelButton, &QPushButton::clicked, &dialog, &QDialog::reject);
+
+    if (dialog.exec() == QDialog::Accepted) {
+        double sizeValue = sizeEdit->text().toDouble();
+        QString sizeUnit = unitCombo->currentText();
+
+        qint64 multiplier = 1;
+        if (sizeUnit == "KB") multiplier = 1024LL;
```
shubhamshahaBMW: I don't think KB split is really needed.
@mfukar  what do you think?

src/mainwindow.cpp
```diff
@@ -1746,6 +1748,19 @@ void MainWindow::on_actionExport_triggered()
     exporterThread->start();
 }
 
+//call for spliting the DLT File
+void MainWindow::on_actionSplitDLTFile_triggered(){
```
shubhamshahaBMW, line 2243: please handle below cases 
1. After running DLT-Viewer first time if we trigger file split W/O opening any file, it causes a crash. We can give error popup saying no open file.
2. start DLT -> connect to hw for live logs -> try to split the file (while connected/disconnected). -> split pop-up does not appear. We can give error popup saying save the file first.

src/filespliting.cpp
mfukar: Yeah perhaps it's not necessary at this point.

src/filespliting.cpp
Renu-Priya411: Data loss is calculated by comparing the size of the original DLT file with the total size of the generated split files. The difference between these two values is reported as lost bytes.[split_loss_stats.csv](https://jira.cc.bmwgroup.net/secure/attachment/12130171/12130171_split_loss_stats.csv) 

This indicates how many bytes from the source file did not make it into the split output, typically due to early parser termination, truncated input, or a write failure.

Separate tail-drop value to identify whether the split stopped at an incomplete message near the end of the file is also tracked.

PFA for reference
[split_loss_stats.csv](https://github.com/user-attachments/files/28629377/split_loss_stats.csv)

src/filespliting.cpp
Renu-Priya411: Progress bar implemented

src/filespliting.cpp
Renu-Priya411: redundant checks removed

src/filespliting.cpp
Renu-Priya411: KB Split removed

src/mainwindow.ui
```diff
@@ -440,6 +441,73 @@
...
+              <property name="text">
+               <string>Desc.</string>
+              </property>
+             </item>
+            </widget>
+           </item>
+          </layout>
+         </item>
+         <item>
+          <widget class="QTreeView" name="exploreView">
```
shubhamshahaBMW: Image

The widget is appearing. some UI issue.

src/filespliting.cpp
```diff
@@ -0,0 +1,245 @@
...
+    m_file = file;
+}
+
+
+//The function is triggered when Split DLT File is clicked which is under File Menu.
+//A dialog box asking for the splitting size is opened where the user can specify the size in MB or GB.
+//The size is casted toqint64 and them parsed to splitOutputFile function.
+//The destination path can also be selected by user which will also be parsend to splitOutFile function.
+
+void FileSpliting::splitDLTFile_triggered(QFile &file,QStringList path){
```
shubhamshahaBMW: unused parameter QFile &file

shubhamshahaBMW (changes requested):

src/mainwindow.ui
Renu-Priya411: Issue resolved

src/filespliting.cpp
Renu-Priya411: parameter removed.

shubhamshahaBMW (approved):

PR #702, opened 2025-07-26
Fix release pipeline

vifactor:


vifactor:
ubuntu 20.04 runner is removed from CI as of July 2025. It is impossible to use it anymore.

vifactor:
macOS and Win artifacts are produced and uploaded: https://github.com/COVESA/dlt-viewer/releases/tag/2.28.1-beta9

bunty95 (changes requested):
changing scripts/linux/installUbuntu20.sh to scripts/linux/installUbuntu22and24.sh

for both build environment, and project will fix the issue. Long term agree, there should only be 1 pipeline for PR and release

alexmucde (approved):

hannesa2 (changes requested):
The Ubuntu artifacts are missing.
Any chance to add it too ? 
image

hannesa2:
I see now the Ubuntu artifacts 
image
but I can't verify, if artifact is working properly

hannesa2:
Unfortunately the Ubuntu release job still runs into an error https://github.com/COVESA/dlt-viewer/actions/runs/16596330956/job/46943762195

hannesa2:
Sorry, was it 05da37900 ? It will not build on conflicts
image
We can reset is back with ease

vifactor:
> The Ubuntu artifacts are missing. Any chance to add it too ?

I've decided not to invest further time in this at the moment, as the required effort seems disproportionate to the potential value. Previously, the Ubuntu artifacts were provided as AppImage bundles, which offer good cross-distro compatibility. However, given that the majority of users (likely around 90%) are on Debian-based distributions, a .deb package remains a practical solution. The existing packaging script still functions with version 2.28, though it does require manual execution. Additionally, most Linux users are typically comfortable building the application from source if necessary.

hannesa2:
At least in my point of view, the main job of an release job is to provide a running artifact. 
I at least (and I guess I'm not alone) will not waste time by build a tool, I just want to use it.
But I'm a MacOS user, and I've want I want.

I will reset it to 2.28.1-beta10

vifactor:
> At least in my point of view, the main job of an release job is to provide a running artifact.

I understand and agree, but making linuxdeployqt-based packaging work on newer ubuntu after it was broken for some time is not a trivial task. So our choice for 2.28 release is to:
- not provide any artifacts for mac and win until ubuntu appimage artifact is fixed again
- provide working artifacts for mac and win now as 2.28.1 version and maybe fix ubuntu for 2.28.2 (or maybe never, if nobody asks or helps)

.github/workflows/Release.yml
```diff
@@ -120,9 +120,9 @@ jobs:
           fetch-depth: 0
           submodules: true
       - name: install build environment
-        run: scripts/linux/installUbuntu20.sh
+        run: scripts/linux/buildUbuntu22and24.sh
       - name: Build project
-        run: scripts/linux/buildUbuntu20.sh
+        run: scripts/linux/installUbuntu22and24.sh
```
hannesa2, line 125: In pullrequst it's vice versa
```
      - name: install build environment
        run: scripts/linux/installUbuntu22and24.sh
      - name: Build project
        run: scripts/linux/buildUbuntu22and24.sh
```
first install an d then build, but here it's now like this 
image

.github/workflows/BuildPR.yml
```diff
@@ -53,53 +44,16 @@ jobs:
           MACOS_CERTIFICATE_PWD: ${{ secrets.P12_PASSWORD }}
           MACOS_CERTIFICATE_NAME: ${{ secrets.APPLE_CERTIFICATE_NAME }}
           MACOS_CI_KEYCHAIN_PWD: ${{ secrets.KEYCHAIN_PASSWORD }}
-        run: |
-          # Turn our base64-encoded certificate back to a regular .p12 file
```
hannesa2, line 57: This pullrequest there is too much stuff, which should be not here.
It touches Windows and MacOS build.
For me, this is mess

I guess the pullrequest should rebased to branch "2.28.x" but doing this, I see a lot of unrelated commits (no feat, I will not push)
image
and I guess these are the only related one 
image
am I right ?

I will make a new pull request and keep this untouched

hannesa2:
> * provide working artifacts for mac and win now as 2.28.1 version and maybe fix ubuntu for 2.28.2 (or maybe never, if nobody asks or helps)

Yes, this would be my suggestion too

hannesa2:
I close this, because successor #708 works now properly and comes with a clear history

PR #706, opened 2025-07-28
test_dltctrlmsg: For some reasons using chars fails with int on some …

### LINKED ISSUE #705: New test fails on Big Endian platforms.
[----------] 1 test from CtrlPayload
[ RUN      ] CtrlPayload.parse
./qdlt/tests/test_dltctrlmsg.cpp:25: Failure
Expected equality of these values:
  ctx.logLevel
    Which is: '\xFF' (-1)
  '\xff'
    Which is: '\xFF' (255)

./qdlt/tests/test_dltctrlmsg.cpp:26: Failure
Expected equality of these values:
  ctx.traceStatus
    Which is: '\xFF' (-1)
  '\xff'
    Which is: '\xFF' (255)

[  FAILED  ] CtrlPayload.parse (0 ms)
[----------] 1 test from CtrlPayload (0 ms total)

Introduced by @vifactor in https://github.com/COVESA/dlt-viewer/commit/63c4f8cedee7b216602fbdb97ba6a47ca34f7315

I don't yet know if this test exposes a real qdlt bug in library, or just a test that needs adjoustment for Big Endian platforms...

LocutusOfBorg:
…architectures (such as arm* ppc64el, s390x)

Closes: #705


qdlt/tests/test_dltctrlmsg.cpp
```diff
@@ -22,7 +22,7 @@ TEST(CtrlPayload, parse) {
 
     auto ctx = app.ctxs[0];
     EXPECT_EQ(ctx.id, "MGR");
-    EXPECT_EQ(ctx.logLevel, '\xff');
-    EXPECT_EQ(ctx.traceStatus, '\xff');
+    EXPECT_EQ(ctx.logLevel, -1);
```
vifactor, line 25: I think, the problem is not here. I guess the byte layout generated by dlt-daemon on a big endian platform for the same message will be different, i.e. `data`-content for the same logLevel and traceStatus will be somewhat altered there.

Question: are you developing dlt-viewer on big-endian platform? If you are just using it there (and not developing), I do not think the tests matter.

LocutusOfBorg: I am building dlt-viewer on Debian, so for both LE and BE architectures, and I would like to execute tests for all archs, to avoid missing regressions...
This is the arch lists:
https://buildd.debian.org/status/package.php?p=dlt-viewer&suite=experimental

vifactor: tbh, I do not completely understand why it fails on BE. `\ff` is decimal -1 for int8_t type. Do you understand the fix?

LocutusOfBorg: not at all :)

RubenGarcia: On some platforms, char is signed and in some it is unsigned. You may have a weird cast -1 => 255 and then another cast to int, where the final comparision -1 == 255 fails..

RubenGarcia: On other platforms, chars are 16 bit, so comparing 0xff to -1 will also fail.

vifactor: @RubenGarcia thanks for the info. Do you have an idea how the test can be rewritten so that it remains valid both on BE and LE platforms?

RubenGarcia: The important thing is that you write down the types of the data that you receive, and understand the conceptual data types.
If the documentation says the thing is supposed to be an unsigned byte, then cast it to uint8_t and only compare it to 255. 

I do not like EXPECT_EQ(ctx.logLevel, '\xff'); because that uses either char or wchar_t which you have no control over.

I also do not like     EXPECT_EQ(ctx.logLevel, -1); because you don't have control over signedness.

Go to the documentation, verify what exact type ctx.logLevel is, and then use
EXPECT_EQ((type)ctx.logLevel, (type)(\<value that the documentation says should be there\>));
It would be best if type is platform agnostic, but since you cast both sides, it should work even if the type changes per-platform.

Don't remove     EXPECT_EQ(ctx.traceStatus, '\xff');, but change it to what the documentation says it should have.

RubenGarcia (approved):

RubenGarcia (changes requested):
Do not remove expect_eq lines and make sure that they match the documentation as requested.

PR #712, opened 2025-08-01
Fix Noble Debian Packaging.

alexmucde:
Remove Focal Debian Configuration.


vifactor:
@alexmucde is there any specific reason why you decided to place those OSS plugins into /usr/share instead of /user/bin? this actually broke their discoverability. It is of course possible to fix, but I have hard time to come up with nice solution

alexmucde:
@vifactor 

It uses the Path

#ifndef PLUGIN_INSTALLATION_PATH
#define PLUGIN_INSTALLATION_PATH ""
#endif

It was set before with qmake. We must set it also with cmake.

In Windows it must be "plugins" and in Linux "/usr/share/dlt-viewer/plugins"

Or i hard code it in the source code dependent on Linux or Windows? MacOs must be also considered?

PR #716, opened 2025-08-05
Fix Debian Build Plugins Path.

alexmucde:


alexmucde:
@vifactor This should work now, the other PR was not working

Issue #701, opened 2025-07-25
Keep single pipeline for PR check and release


vifactor:
The release and pull request (PR) pipelines are maintained as two separate setups. As a result, the regular CI process does not validate the functionality of the release pipeline. This disconnect means that issues within the release pipeline often go unnoticed until the actual release occurs—typically only twice a year—which is far too late for timely resolution.

This might help avoiding issues like this: https://github.com/COVESA/dlt-viewer/discussions/698

PR #719, opened 2025-08-06
Keep single pipeline definition for PR and release

vifactor:
https://github.com/COVESA/dlt-viewer/issues/701


vifactor:
This almost works, except that it somehow misses x86 Mac artifact, although the corresponding jobs are there...

.github/workflows/BuildPR.yml
```diff
@@ -211,5 +213,67 @@ jobs:
       with:
         name: DLTViewer-macos${{ matrix.osversion }}-qt6
         path: build/dist/DLTViewer*.tgz
+  
+  createRelease:
+    name: Create release
+    runs-on: ubuntu-24.04
+    needs: [ buildMac, buildMacOSQt6, buildUbuntu, buildWindows ]
+    if: startsWith(github.ref, 'refs/tags/')
```
hannesa2, line 223: Putting all in one big file and run it with an if condition does not remove any boiler plate code.
Where is the benefit ?

vifactor: having one pipeline file instead of two (which must be kept in sync) has obvious benefits IMO. last release showed why it was bad.

.github/workflows/BuildPR.yml
```diff
@@ -211,5 +213,67 @@ jobs:
...
+      # - name: Download DLT Linux artifacts
+      #   uses: actions/download-artifact@v4
+      #   with:
+      #     name: DLT-Linux
+      #   continue-on-error: true
+        
+      - name: Download DLT macOS artifacts arm64
+        uses: actions/download-artifact@v4
+        with:
+          name: DLTViewer-macos15-qt6
```
bunty95, line 241: This will only create a release for macOS 15 on arm64. If this is the use case, then remove macOS 14 from the qt6 build area.

vifactor: IDK, I do not see an issue having the build for macos-14 checked continuously in the pipeline on a pull request, but not releasing bundle for it. This was the case before this PR. Since I am not a mac user, I even do not know if there is a difference between macos14 and macos15 bundles or they are exchangeable. If you think it makes sense, feel free to add macos-14 release artifact upload

bunty95:
@vifactor issue is corrected. macOS13 and macOS15 artifacts upload

PR #723, opened 2025-08-18
added support for DLTv2 when reading raw files

VRafaelKonstantinou:
I made the assumption that a file can either be DLT or DLTv2. No support for mixed files.

On the mainwindow.cpp when the user import a raw file i check the first message header if it is a dltv2 message.
If it is not then the old existing functionality is executed.
If it is a DLTv2 then the new code is run

in dltv2_file_read_raw
I call dltv2_file_read_header_raw to parse the header. I parse the file for the base header optional header and extended header.
I use the existing DLT function to copy the data

dlt_file_check_version
return the first byte of the file. in that byte, bits 5-7 contain the version if it is a dltv2 message.

storage header
Even though the specification allows dynamic length for the ECU ID i fixed it to ECU1 as it is done for DLT and also to simplify the implementation


src/mainwindow.cpp
```diff
@@ -1424,7 +1424,7 @@ void MainWindow::on_action_menuFile_Import_DLT_Stream_triggered()
 {
     QString fileName = QFileDialog::getOpenFileName(this,
         tr("Import DLT Stream"), workingDirectory.getDltDirectory(), tr("DLT Stream file (*.*)"));
-
+    int version=0;
```
vifactor: would you mind moving this declaration closer to usage?

src/mainwindow.cpp
```diff
@@ -1444,10 +1444,22 @@ void MainWindow::on_action_menuFile_Import_DLT_Stream_triggered()
...
+            outputfile.write((char*)importfile.msg.databuffer,importfile.msg.datasize);
+        }
+    }else if (version==2){
+        /*DLTv2*/
+        while (dltv2_file_read_raw(&importfile,false,0)>=0)
+        {   
+            outputfile.write((char*)importfile.msg.headerbuffer,importfile.msg.headersize);
+            outputfile.write((char*)importfile.msg.databuffer,importfile.msg.datasize);
+        }
     }
```
vifactor: this block can be squashed like this:
```
auto dltReadFunc = (version == 2)  ? dltv2_file_read_raw : dlt_file_read_raw;
while (dltReadFunc(&importfile,false,0)>=0)
        {   
            outputfile.write((char*)importfile.msg.headerbuffer,importfile.msg.headersize);
            outputfile.write((char*)importfile.msg.databuffer,importfile.msg.datasize);
        }
```

qdlt/dlt_common.h
```diff
@@ -355,6 +355,19 @@ typedef struct
...
 } PACKED DltStandardHeaderExtra;
 
+/**
+ * The structure of the DLT file storage header. This header is used before each stored DLT message.
+ */
+typedef struct
+{
+    char pattern[4];		/**< This pattern should be DLT0x01 */
+    uint32_t nanoseconds;			    /**< nanoseconds since 1.1.1970 */
+    uint32_t seconds;            /**< secocnds */
```
vifactor:
```suggestion
    uint32_t seconds;            /**< seconds */
```

qdlt/dlt_common.c
```diff
@@ -1946,6 +2188,19 @@ int dlt_file_free(DltFile *file,int verbose)
...
 }
 
+int dlt_file_check_version(DltFile *file,int verbose)
+{   
+    PRINT_FUNCTION_VERBOSE(verbose);
+    char first_char =fgetc(file->handle);
+    if (first_char != EOF) {
+        rewind(file->handle);
+    } else {
+        printf("File is empty\n");
```
vifactor: it is bad that low level parsing function directly prints something to the console

vifactor (commented):
thanks for the contribution. Could you please add a bit of description to the PR, explaining what has been done here.
I am also curious if the code you added to dlt_common-files is the one from dlt-daemon or it is your custom implementation?

qdlt/dlt_common.c
VRafaelKonstantinou: Fair point. forgot to remove debug print.

src/mainwindow.cpp
VRafaelKonstantinou: done

vifactor (approved):
Would be great to add some basic unit tests for this functionality, otherwise looks ok to me.

PR #726, opened 2025-08-28
qdlt: Add some basic docs

mfukar:
Add some basic descriptions for the headers in qdlt/ so that some relationships are clear in generated graphs.

This is not complete by far, most functions miss documentation and there are no statements of intent anywhere.


qdlt/qdltconnection.h
vifactor, line 1: If unused, pleease remove

vifactor (commented):
This is a good analysis but comments looks often like a notes for yourself. If you intend to merge the MR, please keep only comments where you know what function/class does. If role of a code is not clear, just don't write a comment.

qdlt/qdltconnection.h
mfukar: No I'm not going to do that here. I am not familiar with the intent of the current developers and whether this is some work-in-progress.

mfukar:
> This is a good analysis but comments looks often like a notes for yourself. If you intend to merge the MR, please keep only comments where you know what function/class does. If role of a code is not clear, just don't write a comment.

I can keep working on this change to find out concrete information and have it reflected in the comments you are referring to, and I can also remove them instead, but I think it would be preferable by everyone if instead those notes were replaced by actual facts. Most of them have to do with intent which I am not aware of and I would prefer if the team helped me understand those pieces - and improve the PR accordingly.

PR #729, opened 2025-09-10
Added some documentation for searchDialog

Renu-Priya411:
Doxygen Documentation for searchDialog.cpp

### DIFF
--- a/src/searchdialog.h
+++ b/src/searchdialog.h
@@ -43,28 +43,76 @@ class SearchDialog;
 class QDltFile;
 class QDltPluginManager;
 
+/**
+ * @class SearchDialog
+ * @brief Provides a dialog for searching messages in DLT Viewer.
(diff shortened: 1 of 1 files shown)


src/searchdialog.cpp
```diff
@@ -29,10 +29,13 @@
 #include <QProgressDialog>
 #include <QSignalBlocker>
 
-
+/**
```
vifactor: docstring comments are normally placed in header and not in implementation. Also formatting in many places git broken...

Renu-Priya411: may i know what does formatting refer to ?

vifactor: this is what I see:
image

in any case, if you think that those docstrings are useful, move them to the corresponding header, otherwise it does not make any sense

Renu-Priya411:
Hi @mfukar. The commits are squashed. Thanks

Issue #200, opened 2021-04-23
Export search results into DLT file


alexmucde:
Export all found DLT messages of a search result into a DLT file

alexmucde:
Any contributer available to work on this?

PR #731, opened 2025-09-24
Export search results into DLT file

Pavithra-aa-Anand:
Export all found DLT messages of a search result into a DLT file and "Export in DLT Format" option is in the context menu of search result.

https://github.com/COVESA/dlt-viewer/issues/200

Signed-off by : Pavithra Anand <Pavithra.AA.Anand@bti.bmwgroup.com>

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1570,9 +1570,6 @@ void MainWindow::exportSelection(bool ascii = true,bool file = false,QDltExporte
     disconnect(&exporter,SIGNAL(clipboard(QString)),this,SLOT(clipboard(QString)));
 }
 
-/* Exports search results from the search table view to clipboard or file in various formats.
-   For clipboard operations: uses selected rows only.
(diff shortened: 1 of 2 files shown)


src/mainwindow.cpp
```diff
@@ -1570,15 +1570,34 @@ void MainWindow::exportSelection(bool ascii = true,bool file = false,QDltExporte
     disconnect(&exporter,SIGNAL(clipboard(QString)),this,SLOT(clipboard(QString)));
 }
 
-void MainWindow::exportSelection_searchTable(QDltExporter::DltExportFormat format = QDltExporter::FormatClipboard)
+/* Exports search results from the search table view to clipboard or file in various formats.
```
vifactor: please move such comments to header file

vifactor (approved):

src/mainwindow.cpp
Pavithra-aa-Anand: It's done.

PR #751, opened 2025-12-09
Feature/check for updates

Renu-Priya411:
An automatic pop-up for newer version of DLT Viewer once in 3 months.

### DIFF
--- a/src/mainwindow.h
+++ b/src/mainwindow.h
@@ -134,6 +134,10 @@ private:
     WorkingDirectory workingDirectory;
     bool filterIsChanged;
 
+    //Maps to hold the filter values - findFilteredLines() & MarkedMessages
+    QMap<QString, int> filterCountMap;
+    int totalMessages;
+
     /* Status line items */
     QLabel *statusFilename;
     QLabel *statusFileError;
@@ -405,6 +409,10 @@ private:
                                const EcuItem* ecuitem);
 
 
+
+    void findFilteredLines();
+
+
 protected:
     void keyPressEvent ( QKeyEvent * event ) override;
     void dragEnterEvent(QDragEnterEvent *event) override;
@@ -551,6 +559,7 @@ private slots:
     void on_action_menuFilter_Append_Filters_triggered();
     void onactionmenuFilter_SetAllActiveTriggered();
     void onactionmenuFilter_SetAllInactiveTriggered();
+    void on_actionFiltered_Message_Count_triggered();
 
     // Plugin methods
     void on_action_menuPlugin_Hide_triggered();
--- a/qdlt/qdltfilterlist.cpp
+++ b/qdlt/qdltfilterlist.cpp
@@ -129,6 +129,20 @@ QString QDltFilterList::checkMarker(const QDltMsg &msg)
 
 #endif
 
+const QDltFilter* QDltFilterList::matchMarkerFilter(const QDltMsg &msg) const
+{
+    for(int numfilter=0;numfilter<mfilters.size();numfilter++)
+    {
+        QDltFilter *filter = mfilters[numfilter];
+        if(filter->match(msg))
+        {
+            return filter;
+        }
+    }
+
+    return nullptr;
+}
+
 bool QDltFilterList::applyRegExString(QDltMsg &msg,QString &text)
 {
     QDltFilter *filter;
--- a/qdlt/qdltfilterlist.h
+++ b/qdlt/qdltfilterlist.h
@@ -94,6 +94,8 @@ public:
     QString checkMarker(const QDltMsg &msg);
 #endif
 
+    const QDltFilter* matchMarkerFilter(const QDltMsg &msg) const;
+
 
 
     //! Check if message matches the filter.
--- a/src/dltfileindexer.cpp
+++ b/src/dltfileindexer.cpp
@@ -11,6 +11,7 @@
 #include <QDir>
 #include <QFileInfo>
 #include <vector>
+#include <limits>
 
 #include "qdltoptmanager.h"
 
@@ -395,6 +396,7 @@ bool DltFileIndexer::indexFilter(QStringList filenames)
 
     // get filter list
     filterList = dltFile->getFilterList();
+    resetMarkerCounts(filterList);
     // clear index filter
     indexFilterList.clear();
     indexFilterListSorted.clear();
@@ -426,6 +428,7 @@ bool DltFileIndexer::indexFilter(QStringList filenames)
     {
         // loading filter index from filter is successful
         qDebug() << "Loaded filter index cache for files" << filenames;
+        computeMarkerCountsFromIndex(filterList, indexFilterList);
         return true;
     }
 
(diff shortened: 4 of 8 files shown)


dlt-viewer
vifactor, line 1: what is this submodule?

src/mainwindow.h
```diff
@@ -162,6 +163,9 @@ class MainWindow : public QMainWindow
     SettingsDialog *settingsDlg;
     QDltSettingsManager *settings;
 
+    //Update Pop Up
+    updateChecker *updChecker;
```
vifactor: - class names shall start with capital case
- please do not save on letters and write full words when naming vars, aka updateChecker
- comment is not correct, because besides popup update checker is the whole engine to do the work

src/settingsdialog.cpp
```diff
@@ -134,6 +166,31 @@ SettingsDialog::~SettingsDialog()
     delete ui;
 }
 
+void SettingsDialog::loadUpdateSettings()
+{
+    QSettings settings("MyCompany", "DLTViewer");
+
+    bool isCustom = settings.value("updateCheck/useCustom", false).toBool();
+    int interval = settings.value("updateCheck/customMonths", 3).toInt();
```
vifactor: 3 is a magical constant which is used in different places, please define it clearly with a good name in a global scope

src/updatechecker.cpp
```diff
@@ -0,0 +1,222 @@
...
+        if (latestVersion > currentVersion) return true;
+        if (latestVersion < currentVersion) return false;
+    }
+    return false;
+}
+
+// -----------------------------
+// Popups
+// -----------------------------
+void updateChecker::showUpdatePopup(const QString &current, const QString &latest)
```
vifactor: popup is usually a very poor UX for such information. Normally new version is reported somewhere in a main window as a non-disturbing message.

dlt-viewer
Renu-Priya411: Not sure how this got added. Have removed it in this commit

src/mainwindow.h
Renu-Priya411: HAve made the required changes. Thanks for the review correction

src/updatechecker.cpp
Renu-Priya411: This pop-up doesnt interrupt everytime dlt viewer is opened. It pops up only when the time line reaches. If suppose its given as a undisturbing pop-up, the users who might need an update will never know that there is an update. Thats why its given in the screen itself

Renu-Priya411:
> is there a way to disable the check completely? Ubuntu users normally stick to version existing in their repos

Yes. When there is a pop-up the user can give ignore as well, which will never pop up again till the timeline reaches. Its completely the user's choice to update or no. We are just intimating the user that there is an update available.

src/updatechecker.cpp
bunty95: echoing what vifactor mentioned, could this not be placed in the bottom bar, near "Version"? 
image
instead of showing a pop-up. That would also remove the need for the 3 month check, as it could poll technically 1/day

Renu-Priya411: If the pop up is in the right bottom corner, there are high chances that the pop up will be ignored by the user.
The main purpose of this feature is to intimate the user that there is a latest update available in DLT Viewer. 

In VS Code the pop up usually comes in the right bottom corner which is ignored by many users most of the time. The same issue will happen in DLT Viewer also. That's why I tried to implement the pop up in the center, so that the response has to be given by the user to continue working.

mfukar: I can offer some input as I was involved in the original request. The intent is mainly to inform about major releases, which AFAIK are planned to ramp up to 3-4 per year, and will generally speaking have both important bug fixes which warrant a minor release but hopefully broader and more useful features. This was motivated because we have a lot of users who are using old(er) versions and frequently complain of problems which are solved in recent releases.

Setting the check interval to an arbitrary timespan, or disable it, I think should be offered. A pop-up once every three months when starting up should not be too intrusive but is certainly more intrusive than an icon in the bottom bar. From my PoV grabbing attention like that is desirable. I also think the latter would be mostly ignored due to its persistence. Some other applications go with some text in the bottom bar (e.g. the Zed editor prints something like "Click here to restart and update Zed", next to checkmark or exclamation icon) which I think less of than the pop-up - 
the argument could be made that too much text in the bottom bar can be distracting (I certainly feel that way even about the existing "Version" string).

Anyway, broadly speaking I would not be opposed to putting this notification in the bottom bar, but if so, I would prefer the text to be quite short.

src/settingsdialog.cpp
```diff
@@ -43,6 +43,8 @@
  #define DAYLIGHT daylight
 #endif
 
+// Define default update check interval in months
+static const int DEFAULT_UPDATE_CHECK_MONTHS = 3;
```
mfukar: Following the earlier comments - maybe it would be best to change the unit here to days, just in case a user wants to have frequent checks.

src/CMakeLists.txt
```diff
@@ -85,21 +85,10 @@ add_executable(dlt-viewer
...
-    injectiondialog.ui
-    jumptodialog.ui
-    mainwindow.ui
-    multiplecontextdialog.ui
-    plugindialog.ui
 
-    searchform.ui
-    settingsdialog.ui
-    searchdialog.ui
+    applicationdialog.ui contextdialog.ui ecudialog.ui exporterdialog.ui fileexplorertab.ui filterdialog.ui injectiondialog.ui jumptodialog.ui mainwindow.ui multiplecontextdialog.ui plugindialog.ui searchdialog.ui searchform.ui settingsdialog.ui
```
shubhamshahaBMW: these changes are necessary?

src/settingsdialog.cpp
Renu-Priya411: Hi @mfukar. Since we are heading to official release, we will take up this change for future releases. Requesting you to approve. Thanks in advance

src/CMakeLists.txt
Renu-Priya411: Addressed

src/updatechecker.cpp
```diff
@@ -0,0 +1,213 @@
...
+    : QObject(parent),
+      updateTimer(nullptr),
+      manager(new QNetworkAccessManager(this))
+{
+}
+
+void UpdateChecker::linkToUrl(){
+
+    QString currentVersion = PACKAGE_VERSION;
+    QNetworkRequest request((QUrl(LatestReleaseApi)));
```
shubhamshahaBMW, line 28: missed to set header setRawHeader(<>)

src/updatechecker.cpp
```diff
@@ -0,0 +1,213 @@
...
+        reply->deleteLater();
+
+               // Parse JSON
+        QJsonDocument doc = QJsonDocument::fromJson(jsonResponse.toUtf8());
+        QJsonObject obj = doc.object();
+
+        QString tag = obj["tag_name"].toString();   // Example: "v3.35.0"
+        QString latestVersion = tag.startsWith('v') ? tag.mid(1) : tag;
+
+        QString releaseUrl = obj["html_url"].toString();
```
shubhamshahaBMW: unused variable.

src/updatechecker.cpp
Renu-Priya411: Addressed. Thanks

src/updatechecker.cpp
Renu-Priya411: Addressed. Thanks

PR #790, opened 2026-03-24
Fix DLT file append failure and resolve path encoding for non-Latin1 characters characters

wobuxie1997:
- Fix file opening issues in environments with non-Latin1 characters (e.g., Chinese) by using toLocal8Bit() instead of toLatin1() for dlt_file_open calls.
- Fix on_tabExplore_filesAppendRequest to correctly handle appending .dlt files by invoking appendDltFile.
- Add support for appending .dlf filter files and improve handling for PCAP/MF4 files in the file explorer's append request.
- Refactor file open and append logic in MainWindow for better consistency and broader file type support.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1466,7 +1466,12 @@ bool MainWindow::openDltFile(QStringList fileNames)
         dlt_file_init(&importfile,0);
 
         /* open DLT stream file */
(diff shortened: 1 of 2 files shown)


shubhamshahaBMW:
Hi @wobuxie1997 ,

Show in explorer is not working in windows.

Apart from this, the changes looks file.

shubhamshahaBMW (approved):

PR #792, opened 2026-03-27
Fix CRLF window performance.

Pavithra-aa-Anand:
- Fix sporadic data mismatch in CRLF window during filter/file changes
- Improve rebuild performance when CRLF window is open - currently causes significant slowdown in main window data processing
- Correct export progress bar timing to appear before success popup


bunty95:
Can this be merged to mainline?

src/crlffilterwindow.cpp
```diff
@@ -747,3 +837,198 @@ void CrlfFilterWindow::closeWindow() {
...
+            messageDataCache[actualPos] = data;
+        }
+        }
+    }
+    
+    return data;
+}
+
+// Check if message cache is valid for current filter state
+bool CrlfFilterWindow::isMessageCacheValid(int currentFilteredCount, int /*totalMessages*/) {
```
shubhamshahaBMW: un-used function. not called anywhere.

src/crlffilterwindow.cpp
shubhamshahaBMW, line 244: **** Something missed from previous PR.

applyColumnSettings() is called twice in same function. L244 & L265

shubhamshahaBMW (changes requested):

src/crlffilterwindow.cpp
Pavithra-aa-Anand: removed.

src/crlffilterwindow.cpp
Pavithra-aa-Anand: yes, second call was redundant. Removed it.

shubhamshahaBMW (approved):

PR #798, opened 2026-04-22
Add offscreen platform plugin to macOS bundle

flappy8023:
## Summary
- add the offscreen Qt platform plugin to the macOS app bundle
- resolve the Qt plugin directory via qmake so the packaging works with the Qt6 framework layout
- fail packaging early when the offscreen plugin is missing

### DIFF
--- a/src/cmake/Darwin.cmake
+++ b/src/cmake/Darwin.cmake
@@ -20,6 +20,35 @@ install(PROGRAMS
   DESTINATION "${DLT_EXECUTABLE_INSTALLATION_PATH}/.."
   COMPONENT dlt_viewer)
 
+# Install Qt offscreen platform plugin for silent/help mode on macOS.
+# dlt-viewer sets QT_QPA_PLATFORM=offscreen for -s/--silent and -h/--help,
+# so the plugin must be present in the standard Qt platform plugin bundle path.
+get_target_property(DLT_QMAKE_LOCATION ${QT_PREFIX}::qmake LOCATION)
+if(DLT_QMAKE_LOCATION)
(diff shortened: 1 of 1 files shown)


bunty95:
@flappy8023  Can a use case for this be provided?
running an executable after the ".app" is signed, will cause GateKeeper to throw a fault, and stop the app from running.
Similar to how "dlt-commander" is on macOS.

flappy8023:
> @flappy8023 Can a use case for this be provided? running an executable after the ".app" is signed, will cause GateKeeper to throw a fault, and stop the app from running. Similar to how "dlt-commander" is on macOS.

Thanks for the feedback.
The main use case for this PR is the existing macOS dlt-viewer -s/--silent export flow. In src/main.cpp, dlt-viewer currently sets QT_QPA_PLATFORM=offscreen for -s/--silent and -h/--help, so that code path requires the offscreen platform plugin on macOS.
I also checked dlt-commander as an alternative. It can handle some basic headless export cases, but for me it was not a full replacement:
- it does not fully match the existing dlt-viewer -s workflow
- and in earlier macOS packaging it still required Qt to be installed on the target machine
So from a user/distribution perspective, this left no reliable self-contained macOS command-line export path for the current dlt-viewer -s use case.
Also, this patch does not modify or run anything after signing. It only installs libqoffscreen.dylib during install/package time, before codesigning.
This PR was AI-assisted, and I validated the packaging locally on macOS. The rebuilt app bundle and regenerated dmg both worked on my side.
If I misunderstood the intended macOS packaging model or the expected usage here, I think it is completely fine to close this PR.

PR #803, opened 2026-05-07
remove the check boxes from filter/plugin view.

shubhamshahaBMW:
* Remove the check boxes 
* Add frequently used sort option to toolbar


mfukar (approved):

shubhamshahaBMW:
No longer valid. Had bugs which are fixed by @Pavithra-aa-Anand in #806

PR #807, opened 2026-05-27
Fix unstable search result handling

gernotmuc:
## Summary
- harden search result table bounds checks
- guard search dialog line edit synchronization when toolbar/dialog fields are missing
- fix search result table column sizing and remove a redundant main window refresh

## Validation
- built locally on Windows with Qt 6 and MSVC

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -790,7 +790,7 @@ void MainWindow::initSearchTable()
 
 
     /* set table size and en */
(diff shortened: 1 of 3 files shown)


gernotmuc:
@alexanderawwenzel could you please review this change when you have a moment?

gernotmuc:
Superseded by #814.

I folded the viewer-side search-stability changes from this PR into the new combined upstream PR so the public search work is reviewed in one place together with the follow-up search hot-path optimizations.

The new review thread is: #814.

PR #808, opened 2026-05-29
Guard plugin selection refresh during config reload

gernotmuc:
## Summary
- suppress plugin-driven selection refresh while a plugin configuration change reloads the current log
- - clear the live selection before forcing the reload path
- - rebuild plugin state through the normal log reload flow after config changes
## Notes
- public port of the validated internal viewer-side fix


gernotmuc:
- core/base PR: #818

PR #810, opened 2026-06-02
Preserve all CTIDs in config tree across GetLogInfo messages

gernotmuc:
Fix EcuTree app merge logic to avoid overwriting APID contexts when multiple GetLogInfo control messages are parsed from a log file. This restores full APID->CTID trees used by Config tab and Save IDs as CSV export.

Add a regression test covering repeated GetLogInfo messages for the same APID with different CTIDs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,164 +1,51 @@
-# SPDX-License-Identifier: MPL-2.0
-#
-# Copyright (C) 2016, Jack S. Smith
-#
-# This file is part of COVESA DLT-Viewer project.
-#
-# This Source Code Form is subject to the terms of the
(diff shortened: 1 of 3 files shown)


gernotmuc:
Quick context for reviewers:

- Root cause: `EcuTree::add()` overwrote APID entries when multiple GetLogInfo control messages for the same APID were parsed, so previously collected CTIDs were lost.
- Fix: merge into existing APID node instead of replacing it.
- Added regression test: `EcuTree.PreservesContextsFromMultipleGetLogInfoMessagesForSameApp`.

Validation status:
- Local viewer build: OK (both COVESA clean and BMW variant used for packaging).
- Unit test execution: not run in this environment because GTest runtime was not available here.

@lutzbichler could you please review?

PR #811, opened 2026-06-02
Fix/config ctid merge saveidscsv clean

gernotmuc:


gernotmuc:
Superseding fix after path cleanup: this PR contains only the intended files (`src/ecutree.cpp`, `qdlt/tests/CMakeLists.txt`, `qdlt/tests/test_ecutree.cpp`).

Reviewer context correction: this should go to **Alexander Wenzel** (not Lutz).

gernotmuc:
Reviewer context correction: Julian Tiedeken should be the main reviewer/contact for this PR.

All checks are green and the PR is ready for review.

mfukar (changes requested):
Hi Gernot, long time no see :D

Would you please add some descriptions of the regressions fixed & tested and a brief PR description as well?

gernotmuc:
@mfukar thanks, adding the missing context here.

## Summary
This PR fixes a regression in the `EcuTree` merge path for repeated `GetLogInfo` responses.

Before this change, when the same ECU/application was reported in multiple control messages, the later message recreated the application entry and overwrote the previously collected contexts. In practice that could leave only the last CTID visible in the merged tree, which then also affected consumers of that merged APID/CTID data such as the save-IDs/export path.

## Regressions Fixed
- Repeated `GetLogInfo` updates for the same ECU/APID no longer drop previously collected CTIDs.
- The application description is still refreshed from the incoming message, but the already known contexts are preserved and extended.
- The merged ECU/application/context view now stays complete when context information arrives in multiple chunks instead of one single message.

## Testing
- Added a dedicated regression test `EcuTree.PreservesContextsFromMultipleGetLogInfoMessagesForSameApp` in `qdlt/tests/test_ecutree.cpp`.
- The test feeds two `GetLogInfo` payloads for the same ECU/app with different contexts and verifies that both `CTX1` and `CTX2` are still present afterwards with the expected descriptions.
- Updated `qdlt/tests/CMakeLists.txt` so `test_ecutree.cpp` and `src/ecutree.cpp` are built into the `test_qdlt` target and executed in CI.
- Fixed the test code for broader compiler compatibility by replacing `std::map::contains(...)` with `find(...) != end()`, which is why the small follow-up test commit is part of this PR.
- All PR checks are green again after the regression-test integration and compatibility fix.

gernotmuc:
- core/base PR: #818

PR #812, opened 2026-06-04
Plugin pane filename in sync with plugin widget filename

Pavithra-aa-Anand:
Ensure filename is visible and consistent in plugin pane and dialog by syncing displayed filename in pane.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -6795,6 +6795,11 @@ void MainWindow::on_action_menuPlugin_Edit_triggered() {
         dlg.setPluginInterfaceVersion(item->getPluginInterfaceVersion());
         dlg.setFilename(item->getFilename());
         dlg.setMode(item->getMode());
+        // Keep dialog filename in sync with the value currently shown in plugin view.
+        const QString displayedFilename = item->text(2);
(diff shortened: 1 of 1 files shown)


shubhamshahaBMW (approved):

src/mainwindow.cpp
```diff
@@ -6795,6 +6795,11 @@ void MainWindow::on_action_menuPlugin_Edit_triggered() {
         dlg.setPluginInterfaceVersion(item->getPluginInterfaceVersion());
         dlg.setFilename(item->getFilename());
```
shubhamshahaBMW, line 6796: Should the dialog setFileName() be called after 6802?

shubhamshahaBMW (changes requested):

src/mainwindow.cpp
Pavithra-aa-Anand: Yes, check if the displayed filename differs from the stored one and updates item->setFilename(displayedFilename) — after the dialog has already been initialized with the old filename.

shubhamshahaBMW (approved):

Renu-Priya411 (approved):

PR #813, opened 2026-06-08
Speed up Find All search in viewer

gernotmuc:
## Summary
- remove the Qt6-specific QtConcurrent Find All path and use the existing sequential search loop again
- avoid decoding search-result columns that do not need decoded payload data
- cache per-search settings/text once per scan and guard the line-edit sync paths

## Why
The Qt6 parallel Find All path regressed large-log search performance noticeably. Reusing the existing sequential path restores the expected behavior and keeps result ordering and UI updates simpler.

## Validation
- compare contains only src/searchdialog.cpp, src/searchdialog.h, and src/searchtablemodel.cpp
- ported exactly from the validated BMW viewer change set for the Qt6 search-speed fix
- benchmark on the investigated large log restored Find All time to about 2:43 with SomeIP enabled and about 0:56 without SomeIP

### DIFF
--- a/src/searchdialog.cpp
+++ b/src/searchdialog.cpp
@@ -31,22 +31,13 @@
 #include <QSignalBlocker>
 #include <QColorDialog>
 #include <QAction>
-#include <QPointer>
-#include <QThreadPool>
-#include <QThread>
 #include <QDebug>
-#include <QtConcurrent/QtConcurrent>
-
-#include <mutex>
 
 SearchDialog::SearchDialog(QWidget *parent) :
     QDialog(parent),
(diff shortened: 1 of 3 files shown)


gernotmuc:
Additional context for reviewers: this is the public port of the already validated BMW Qt6 search-speed fix. The main functional change is removing the Qt6-specific parallel Find All path and reusing the existing sequential path again. On the investigated large log, the validated benchmark restored Find All time to about 2:43 with SomeIP enabled and about 0:56 without SomeIP.

gernotmuc:
- core/base PR: #818

PR #814, opened 2026-06-09
Improve viewer search stability and speed

gernotmuc:
## Summary

This PR combines the public search-stability fix that was previously prepared as #807 with the follow-up viewer search hot-path optimizations, so the remaining viewer-side search changes can be reviewed and merged in one upstream PR.

- fix unstable search result handling in the viewer search dialog / search results model
- cache the stable `QDltMsg::toStringHeader()` suffix and the last formatted local time string instead of rebuilding them for every row
- store the optional time filter as milliseconds since epoch and skip per-message `QDateTime` construction unless a time-range search is active
- reduce unnecessary work in `SearchDialog` and `SearchTableModel` during Find All / result rendering by only decoding when payload data is actually needed and by avoiding redundant refresh paths
- switch the cancel flag in the search dialog to `std::atomic_bool` so UI-driven cancellation remains safe while the search loop polls it

## Notes

- this supersedes the separate public search-stability PR `#807`; its viewer-side changes are included here now
- no changes from public PR `#808` or `#811` are included here
- the head branch is still `gernotmuc:fix/viewer-search-speed-public`; the compare against `COVESA:master` intentionally includes the earlier search-stability commits plus the new hot-path speed changes

## Validation

- `get_errors` is clean on:
  - `qdlt/dltmessagematcher.cpp`
  - `qdlt/dltmessagematcher.h`
  - `qdlt/qdltmsg.cpp`
  - `src/mainwindow.cpp`
  - `src/searchdialog.cpp`
  - `src/searchdialog.h`
  - `src/searchtablemodel.cpp`
- benchmark context from the public Qt6 speed probes:
  - prior public candidate with header-cache only: without SomeIP `0:37 / 1:41`, with SomeIP `3:05 / 4:06`
  - after adding the matcher millisecond-range optimization on the same public line: without SomeIP `0:38 / 1:02`, with SomeIP `3:33 / 3:23`


gernotmuc:
Closing this in favor of #813 to keep the narrower public search PR and avoid overlapping search lines.

PR #815, opened 2026-06-10
Fail safely on malformed DLT filter files

gernotmuc:
## Summary

This change makes malformed DLT filter files fail safely and report the actual parser error to GUI users.

## Regressions Fixed

- Loading a malformed `.dlf` file no longer leaves partially parsed filter entries behind.
- Reloading a default filter directory now skips malformed files instead of treating them like valid filter lists.
- GUI users now get an English warning with the affected file name and parser details both for skipped default filters and for manual `Filter -> Load` failures.

## Implementation

- Make `QDltFilterList::LoadFilter()` atomic by collecting parsed filters in a temporary list and only committing them on successful XML parsing.
- Store the last parser error text, including line and column information, on the filter list.
- Record malformed default filter files during directory load and show a warning dialog after the load finished.
- Reuse the stored parser error text in `Project::LoadFilter()` so manual filter loading shows a detailed error dialog instead of a generic failure message.

## Testing

- Built the isolated branch successfully on Windows with Qt 6.7.2 and MSVC 2019:
  - `cmake -S C:\dlt_devel\01_tmp\covesa-filterfile-pr -B C:\dlt_devel\01_tmp\covesa-filterfile-pr\build\Release -DCMAKE_PREFIX_PATH=C:\Qt\6.7.2\msvc2019_64`
  - `cmake --build C:\dlt_devel\01_tmp\covesa-filterfile-pr\build\Release --target dlt-viewer --config Release`
- Verified editor diagnostics are clean for the touched files.
- Covered the malformed default-filter startup path and the manual `Filter -> Load` failure path in the implementation so both surface the parser details to GUI users while silent mode stays log-only.

### DIFF
--- a/qdlt/qdltdefaultfilter.cpp
+++ b/qdlt/qdltdefaultfilter.cpp
@@ -43,6 +43,9 @@ QDltDefaultFilter::~QDltDefaultFilter()
 
 void QDltDefaultFilter::clear()
 {
+    malformedFilterFiles.clear();
+    malformedFilterErrors.clear();
+
     /* delete all filter list entries */
     QDltFilterList *t;
     foreach(t,defaultFilterList)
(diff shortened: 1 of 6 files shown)


gernotmuc:
- core/base PR: #818

PR #816, opened 2026-06-10
Speed up filter index rebuild hot paths

gernotmuc:
## Summary

This change speeds up the hot path that is exercised when DLT Viewer rebuilds filter indices for large logs with payload-heavy filter files.

The optimization stays inside the existing filter semantics: it does not move plugin decoding, change which filters match, or alter filter precedence for equivalent filters.

This PR is intentionally scoped to the filter-matching performance slice only. The malformed filter file handling from #815 stays separate.

## Regressions Fixed

- Rebuilding filter indices for large logs was spending too much time in repeated QString construction and copying inside QDltFilter::match().
- - Payload-heavy and header-heavy filter sets were evaluating expensive checks too early, which amplified the slowdown during `Filter -> Load` and similar rebuild paths.
- - Large reference-filter scenarios regressed versus the earlier public baseline even though the matching result itself did not change.
## Implementation

- Add const-reference accessors for ECU/APID/CTID and use cached header/payload string accessors in QDltFilter::match() so repeated filter checks can reuse already materialized data instead of rebuilding or copying QStrings on each check.
- - Keep the current upstream header/time hot-path helpers intact and add per-message header/payload reuse inside QDltMsg, with invalidation when the relevant message state changes.
- - Sort active positive and negative filters with a stable cost/specificity heuristic so cheaper and more selective checks run before expensive header/payload matches, while preserving relative order for equally ranked filters.
## Testing

- Built the isolated branch successfully on Windows with Qt 6.7.2 and MSVC 2019:
-   - `cmake -S C:\dlt_devel\01_tmp\covesa-filterhot-pr -B C:\dlt_devel\01_tmp\covesa-filterhot-pr\build\Release -DCMAKE_PREFIX_PATH=C:\Qt\6.7.2\msvc2019_64`
-   - `cmake --build C:\dlt_devel\01_tmp\covesa-filterhot-pr\build\Release --target qdlt dlt-viewer --config Release`
- - Rebased the final four-file diff onto the current public `master` sources and verified editor diagnostics are clean for that rebased change set.
- - Validated the real regression scenario on Windows with a large log and a payload-heavy reference filter file:
-   - Scenario: load `455.all`, delete filter indices, then load `gernot_network.dlf`.
-   - Clean public baseline: `1:21`
-   - Installed public SDK stand: `1:36`
-   - New optimized stand: `1:11`
That keeps the matching behavior intact while recovering 10 seconds against the clean baseline and 25 seconds against the slower installed stand in the measured workflow.


gernotmuc:
Additional reviewer context:

- Scope: this PR only addresses the filter-index rebuild performance regression. The malformed filter-file / message-box handling stays separate in #815.
- - Root cause: the rebuild path was repeatedly materializing and copying QString data inside `QDltFilter::match()`, and expensive header/payload filters were often evaluated too early during `Filter -> Load`.
- - Safety: I kept the plugin decode ordering untouched. The change is limited to cached/reference string reuse in `QDltMsg` and a stable cost/specificity ordering for active positive/negative filters, so the matching behavior stays the same.
- - Validation: rebuilt on Windows with Qt 6.7.2 + MSVC 2019, rebased the final four-file diff onto current public `master`, and reran the real repro with `455.all` plus `gernot_network.dlf`.
- - Measured result: clean public baseline `1:21`, installed public SDK stand `1:36`, patched stand `1:11`.

gernotmuc:
Closing this in favor of the newer public filter line in #817 so only one public filter-performance PR remains open. The malformed-filter handling in #815 stays a separate additive bugfix line.

PR #817, opened 2026-06-11
Short-circuit filter rejects before decode

gernotmuc:
## Summary

- add a metadata-only prefilter path in `QDltFilterList`
- decide `Match` / `Reject` / `NeedsDecode` before the normal filter matcher runs
- skip decoder work for messages that are already definite rejects when no viewer plugin needs decoded message callbacks

## Why

Applying a loaded filter file currently pushes every message through the decode path before the filter decision is known. For common filters that only look at ECU/APID/CTID/message id/type/level metadata, that work is unnecessary.

This change keeps header, payload, and regexp-on-decoded-text filters on the existing full matcher path, but short-circuits the obvious metadata-only rejects early.

## Validation

- `get_errors` clean on `qdlt/qdltfilterlist.h`, `qdlt/qdltfilterlist.cpp`, and `src/dltfileindexerthread.cpp`
- clean Qt6/MSVC build succeeded in `C:\dlt_devel\01_tmp\covesa-filter-prefilter-pr-20260611\build\Release`
- user stopwatch results on the preferred stand `C:\dlt_devel\01_tmp\qt6_covesa_filterhot_20260611_prefilter`:
  - `gernot_network.dlf`: `1:11`
  - `Lifecycle_gernot`: `1:06`
  - `message_buffer_overflow`: `0:59`

## Notes

- this is the validated winning step only; the later sync-message reuse experiment was intentionally not included because it regressed the first real filter case

### DIFF
--- a/qdlt/qdltfilterlist.cpp
+++ b/qdlt/qdltfilterlist.cpp
@@ -33,6 +33,109 @@ extern "C"
 #include "dlt_common.h"
 }
 
+namespace {
+enum class FilterMatchState { Match, Reject, NeedsDecode };
+
+bool matchesMetadataOnly(const QDltFilter *filter, const QDltMsg &msg)
+{
+    const QString msgEcuid = msg.getEcuid();
+    const QString msgApid = msg.getApid();
+    const QString msgCtid = msg.getCtid();
+
+    if(filter->enableEcuid && (msgEcuid != filter->ecuid))
+    {
+        return false;
+    }
+
+    if(filter->enableRegexp_Appid)
+    {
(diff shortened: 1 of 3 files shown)


gernotmuc:
Additional reviewer context for the public filter line:

- `#815` is the additive malformed-filter bugfix line.
- `#817` is the filter-performance line.
- If you prefer a straight merge sequence, land `#815` first and then `#817`.
- There is no semantic dependency from `#817` onto `#815`; `#815` improves malformed filter loading/warnings, while `#817` only short-circuits metadata-only rejects before decode.

That keeps the public series linear without mixing the bugfix and performance topics into one PR.

gernotmuc:
- core/base PR: #818

PR #821, opened 2026-06-24
Feature/delink UI models

Renu-Priya411:
Delinking the UI thread from search, live logging, filtering and cache. This gives better UI performance


Pavithra-aa-Anand:
Enhancements:

- Move single-step and fallback search off UI thread.
- Live logging: Move socket readyRead parsing/extraction/pre-processing to workers(rowupdate).
- Shifts decode and filter checks for logging-only-filtered flow to worker side.

Improvement:
Phase 4: Decode + filter check in logging-only flow still runs in main thread.

src/filterthreadworker.cpp
```diff
@@ -0,0 +1,49 @@
+#include "filterthreadworker.h"
+
+#include "decodemanager.h"
```
shubhamshahaBMW, line 3: class not used anywhere.

src/crlffilterwindow.cpp
```diff
@@ -718,47 +620,31 @@ void CrlfFilterWindow::rebuildCrlfModel() {
...
+            }
             updateMessageCount(0);
             lastFilteredMessageCount = totalFilteredMessages;
             return;
         }
         
         int actualPos = dltFile->getMsgFilterPos(i);
-        if (actualPos >= 0) {
-            // Check CRLF directly on this filtered message and suppress index building since we already built it above
-            QVariantList messageData = this->extractMessageData(i, true);
```
shubhamshahaBMW, line 739: extractMessageData is nowhere called now, definition can be removed.

src/indexthreadworker.cpp
```diff
@@ -0,0 +1,361 @@
...
+
+namespace {
+
+struct ScannedMessage
+{
+    int globalIndex{-1};
+    QByteArray rawMessage;
+};
+
+QVector<qint64> scanNewMessagePositions(QFile &file, qint64 lastIndexedPosition)
```
shubhamshahaBMW, line 17: this might be causing issue in live logging with huge load.

Image seeing such messages while testing the PR.

shubhamshahaBMW (changes requested):

qdlt/qdltfile.cpp
```diff
@@ -621,6 +647,22 @@ QString QDltFile::getFileName(int num)
...
+    if(num<0 || num>=files.size())
+    {
+        return -1;
+    }
+
+    const QVector<qint64> &indexAll = files[num]->indexAll;
+    if(indexAll.isEmpty())
+    {
+        return -1;
+    }
```
mfukar, line 661: I would expect this to search in the file and provide the offset, rather than returning an invalid one. Should simplify the logic in the caller(s)

qdlt/qdltfile.h
mfukar, line 1: A general comment here, descriptions need a bit more information. For example, `appendDltIndices`: where are the message offsets from? what is the meaning of the parameters, preconditions, etc.

src/indexthreadworker.cpp
mfukar, line 17: Same comment as in qdltfile.h - need more precise info on what these functions intend to do, and why

src/indexthreadworker.cpp
```diff
@@ -0,0 +1,361 @@
...
+
+        if(!file.seek(start))
+        {
+            continue;
+        }
+
+        ScannedMessage message;
+        message.globalIndex = fileInput.baseGlobalIndex + fileInput.existingMessageCount + idx;
+        message.rawMessage = file.read(end - start);
+        messages.push_back(message);
```
mfukar, line 217: if you've already reserved memory in the vector (also, please use standard containers instead of Qt unless where absolutely necessary), `emplace*` migh be a better choice

mfukar (commented):
Just leaving some comments for now

qdlt/qdltfile.h
Renu-Priya411: Hello @mfukar 
This task has been splitted into different phases and is present as PR in the board. The PR numbers are 823, 825,826 and 827. Please review those PRs as well. I have made improvements wrt to this in those PRs. Requesting to check them

Renu-Priya411:
Since this implementation was splitted into separate phases, the PRs are raised for it. Hence closing this PR

Renu-Priya411:
Since this implementation was splitted into separate phases, the PRs are raised for it. Hence closing this PR

PR #823, opened 2026-07-21
Phase 5: filter worker, incremental proxy index, and lightweight CRLF…

Renu-Priya411:
… view

Phase 5: filter worker, incremental proxy index, and lightweight CRLF view

### DIFF
--- a/src/crlffilterwindow.cpp
+++ b/src/crlffilterwindow.cpp
@@ -443,6 +443,17 @@ void CrlfFilterWindow::rebuildCrlfModel() {
         return;
     }
 
+    // Guards every call path (direct calls from createCrlfWindow()/refreshWindow() included,
+    // not just the debounced timer) against reentrancy via processEvents() pumped below.
+    if (rebuildInProgress) {
+        return;
+    }
+    struct InProgressGuard {
+        bool &flag;
+        explicit InProgressGuard(bool &f) : flag(f) { flag = true; }
+        ~InProgressGuard() { flag = false; }
+    } inProgressGuard(rebuildInProgress);
+
     lastBuildCanceled = false;
 
     if (!dltFile || dltFile->size() == 0) {
@@ -544,9 +555,7 @@ void CrlfFilterWindow::onRebuildTimerTimeout() {
     if (!dltFile || dltFile->size() == 0) {
         onSourceModelReset();
     } else {
-        rebuildInProgress = true;
         rebuildCrlfModel();
-        rebuildInProgress = false;
     }
 }
 
--- a/src/crlffilterwindow.cpp
+++ b/src/crlffilterwindow.cpp
@@ -10,9 +10,6 @@
 #include <QCoreApplication>
 #include <QTimer>
 #include <QFileDialog>
-#include <QStandardItemModel>
-#include <QStandardItem>
-#include <QThread>
 
 #include "crlffilterwindow.h"
 #include "mainwindow.h"
@@ -30,17 +27,15 @@ CrlfFilterWindow::CrlfFilterWindow(QObject* parent) : QObject(parent) {
     dltFile = nullptr;
     pluginManager = nullptr;
     
-    // Initialize debouncing mechanism
     rebuildTimer = new QTimer(this);
     rebuildTimer->setSingleShot(true);
     rebuildTimer->setInterval(500);
     connect(rebuildTimer, &QTimer::timeout, this, &CrlfFilterWindow::onRebuildTimerTimeout);
     
     lastFilteredMessageCount = -1;
-    lastCacheValidCount = -1;
-    bulkCrlfIndexBuilt = false;
     rebuildScheduled = false;
     rebuildInProgress = false;
+    lastBuildCanceled = false;
 }
 
 // Check if a message contains CRLF characters
@@ -99,7 +94,6 @@ QStringList CrlfFilterWindow::createTableHeaders() {
 
 // Creates a single window displaying all CRLF messages
 void CrlfFilterWindow::createCrlfWindow() {
-    // Validate prerequisites
     if (!dltFile || dltFile->size() == 0) {
         QMessageBox::information(nullptr, "No DLT file", "No DLT file is currently loaded.");
         return;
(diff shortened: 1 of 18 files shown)


src/indexrowreferencemodel.cpp
```diff
@@ -0,0 +1,93 @@
+#include "indexrowreferencemodel.h"
+
+IndexRowReferenceModel::IndexRowReferenceModel(QObject *parent)
+    : QAbstractTableModel(parent)
+{
+}
+
+void IndexRowReferenceModel::setSourceModel(QAbstractItemModel *newSourceModel)
```
Pavithra-aa-Anand, line 8: setSourceModel never connects to dataChanged, so decoder-plugin updates to existing rows are invisible to every view backed by this model (CRLF window, all ECU tabs). Messages decoded after the proxy is populated will show stale payload text until the whole model is reset.

 Connect sourceModel->dataChanged and forward it by mapping the stored row indices back to proxy rows and also disconnect from the old model when the source is replaced.

src/mainwindow.cpp
```diff
@@ -7050,7 +7100,6 @@ void MainWindow::showCrlfMessages()
     // Check if CRLF window already exists
     if (crlfFilterWindow) {
         crlfFilterWindow->refreshWindow();
-        crlfFilterWindow->showAndActivate();
```
Pavithra-aa-Anand, line 7103: refreshWindow() hides the window before rebuilding and only re-shows it when rowCount() > 0. If no CRLF rows remain after a filter change the window disappears silently — and because showAndActivate() was removed from showCrlfMessages(), clicking "Show CRLF Messages" again does nothing to recover it.

Re-add crlfFilterWindow->showAndActivate() after refreshWindow() in showCrlfMessages() or show a QMessageBox when the rebuild produces 0 rows.

src/mainwindow.cpp
```diff
@@ -2662,6 +2680,8 @@ void MainWindow::reloadLogFileFinishDefaultFilter()
 
 void MainWindow::reloadLogFile(bool update, bool multithreaded)
 {
+    resetLiveFilterGeneration();
```
Pavithra-aa-Anand, line 2701: resetLiveFilterGeneration() clears the pending queue so old messages are discarded, which is correct. 
However, syncLiveFilterWorkerConfig() is not called here, so the live-filter worker keeps the filter configuration from the previous session. If a new file is loaded (or filters changed just before reload), updateIndex() will enqueue new-file messages against stale filters until reloadLogFileFinishFilter() eventually calls filterUpdate() → syncLiveFilterWorkerConfig().

Call syncLiveFilterWorkerConfig() directly after resetLiveFilterGeneration() inside reloadLogFile().

src/filtergrouplogs.cpp
```diff
@@ -239,6 +291,8 @@ void filtergrouplogs::onTabCloseRequested(int index) {
     mergedTabWidget->removeTab(index);
     mergedTabs.remove(tabKey);
     tabToSelectedIds.remove(widget);
+    ecuTabModels.remove(tabKey);
```
Pavithra-aa-Anand: ecuTabModels.remove(tabKey) removes the pointer from the map but never deletes the object. Models accumulate for the session lifetime since their Qt parent is filtergrouplogs, not the tab widget.

Delete the model when the tab is removed: delete ecuTabModels.take(tabKey);

src/crlffilterwindow.cpp
Pavithra-aa-Anand: createTableHeaders() is never called — the call was removed when QStandardItemModel was replaced with IndexRowReferenceModel, but the function was not. Remove both the declaration and the definition

src/crlffilterwindow.cpp
```diff
@@ -678,109 +462,84 @@ void CrlfFilterWindow::rebuildCrlfModel() {
...
-        // If user cancelled during index building, exit early
-        if (!bulkCrlfIndexBuilt) {
-            updateMessageCount(0);
-            return;
-        }
-    }
-    
-    //Now rebuild the model data
+
+    crlfFilterProxy->setSourceModel(sourceModelOfDLT);
```
Pavithra-aa-Anand: rebuildCrlfModel() calls setSourceModel() followed by setRowReferences(). Each call emits beginResetModel()/endResetModel(), causing the attached QTableView to clear and redraw twice. setSourceModel() also clears sourceRows (see indexrowreferencemodel.cpp:10), which is immediately overwritten by setRowReferences() — so the first reset does nothing useful.

The source model is already set by CrlfFilterWindow::setSourceModel() (line 321) and in createCrlfWindow() (line 118). Calling it again inside every rebuild is redundant.

Remove the crlfFilterProxy->setSourceModel(sourceModelOfDLT) call from rebuildCrlfModel().

src/filtergrouplogs.cpp
```diff
@@ -21,40 +21,149 @@
...
 filtergrouplogs::filtergrouplogs(QObject* parent) : QObject(parent) {
     sourceModelOfDLT = nullptr;
-    ecuIdFilterProxy = nullptr;
     mergedTabWidget = nullptr;
     dltFile = nullptr;
     pluginManager = nullptr;
 }
 
 // Extracts unique ECU IDs from a DLT file
 QStringList filtergrouplogs::extractEcuIds(const QString& dltFilePath) {
```
Pavithra-aa-Anand: Remove the parameter: QStringList filtergrouplogs::extractEcuIds(). Update the corresponding declaration in filtergrouplogs.h:22. The dltFilePath parameter is silently ignored (Q_UNUSED)

src/filtergrouplogs.cpp
```diff
@@ -401,3 +466,24 @@ void filtergrouplogs::setDltFile(QDltFile* file) {
 void filtergrouplogs::setPluginManager(QDltPluginManager* manager) {
     pluginManager = manager;
 }
+
+void filtergrouplogs::onSourceModelChanged()
+{
+    rebuildGroupedIndex();
```
Pavithra-aa-Anand, line 467: onSourceModelChanged() does a full O(N) re-scan on every filter/file/live-data event — even when the ECU grouping window is closed. Because a new filtergrouplogs object is created on every splitLogsEcuid() call and each one stays signal-connected, the cost multiplies: N rows × M past invocations on every model reset.

Add a guard: if(!mergedTabWidget || mergedTabs.isEmpty())
        return;

src/mainwindow.cpp
```diff
@@ -831,6 +839,15 @@ void MainWindow::initFileHandling()
     const bool sortByTimestampEnabled = QDltSettingsManager::getInstance()->value("startup/sortByTimestampEnabled", false).toBool();
     dltIndexer->setSortByTimestampEnabled(filtersEnabled && sortByTimestampEnabled);
 
+        liveFilterWorker = new FilterThreadWorker(this);
```
Pavithra-aa-Anand: check the indentation here

src/dltfileindexerthread.cpp
```diff
@@ -112,32 +117,30 @@ void DltFileIndexerThread::processMessage(QSharedPointer<QDltMsg> &msg, int inde
...
-     (void) pluginManager->decodeMsg(*msg, silentMode);
+     (void) pluginManager->decodeMsg(msg, silentMode);
      }
 
 
-    bool_result = filterList->checkFilter(*msg);
+    bool_result = filterList->checkFilter(msg);
     if ( bool_result == true)
     {
-        if(const QDltFilter *markerFilter = filterList->matchMarkerFilter(*msg); markerFilter != nullptr)
```
shubhamshahaBMW: Image

Marker count is no longer calculated for filters.

src/crlffilterwindow.cpp
```diff
@@ -835,185 +592,6 @@ void CrlfFilterWindow::closeWindow() {
...
-        
-        // Cache the processed data for future use
-        messageDataCache[actualPos] = data;
-    } else {
-        // Fallback: Only process if we know it has CRLF (from bulk index)
-        if (actualPos >= 0 && actualPos < dltFile->size()) {
-            QDltMsg msg;
-            if (dltFile->getMsg(actualPos, msg)) {
-            if (pluginManager) {
-                pluginManager->decodeMsg(msg, true); // Silent mode
```
shubhamshahaBMW, line 907: w/o decodeMsg, CRLF is broken for decoder plugins.

src/mainwindow.cpp
```diff
@@ -6999,34 +7070,14 @@ void MainWindow::filterIndexEnd()
...
+    {
+        QMessageBox::warning(this, "Warning", "No table model is currently available.");
+        return;
+    }
+
     if (qfile.getNumberOfFiles() > 0) {
         filtergrouplogs *filterLogsEcuid = new filtergrouplogs(this);
-        // Get the path of the currently loaded DLT file
-        QString currentFilePath = qfile.getFileName(0);
-        QStringList ecuIds = filterLogsEcuid->extractEcuIds(currentFilePath);
```
shubhamshahaBMW, line 7057: extractEcuIds() is not used anymore.

src/filtergrouplogs.cpp
```diff
@@ -175,50 +260,12 @@ void filtergrouplogs::mergeTabs()
...
-    auto settings = QDltSettingsManager::getInstance();
-    for (int col = 0; col < ecuIdFilterProxy->columnCount(); ++col) {
-        bool show = FieldNames::getColumnShown(static_cast<FieldNames::Fields>(col), settings);
-        mergedView->setColumnHidden(col, !show);
-    }
-    int mergedtabIndex = mergedTabWidget->addTab(mergedView, tabKey);
+    createOrUpdateTab(tabKey, rowsForEcuSet(QSet<QString>(selectedIds.begin(), selectedIds.end())));
+    int mergedtabIndex = mergedTabWidget->indexOf(mergedTabs.value(tabKey));
     mergedTabWidget->setCurrentIndex(mergedtabIndex);
     indexofMergedTabs[mergedtabIndex] = tabKey;
```
shubhamshahaBMW: indexofMergedTabs is updated but never used.

shubhamshahaBMW (changes requested):

src/crlffilterwindow.cpp
Renu-Priya411: fixed

src/filtergrouplogs.cpp
Renu-Priya411: fixed

src/mainwindow.cpp
Renu-Priya411: fixed

Pavithra-aa-Anand (approved):

src/filterthreadworker.cpp
```diff
@@ -0,0 +1,144 @@
...
+    QVector<qint64> matchedIndices;
+    matchedIndices.reserve(kBatchSize);
+
+    quint64 activeGeneration = 0;
+    bool hasActiveGeneration = false;
+
+    while(true)
+    {
+        PendingMessage pendingMessage;
+        QDltFilterList filterList;
```
shubhamshahaBMW: filterList = currentFilterList performs a deep copy of the complete QDltFilterList for every dequeued message. This allocates and copies every QDltFilter repeatedly, even when the filter configuration has not changed, and it also happens when filtering is disabled.

Using currentFilterList directly without synchronization would introduce a data race because setFilterConfiguration() replaces and deletes its filter objects under queueMutex. Holding the mutex during checkFilter() would avoid the race but block message enqueueing and configuration updates.

Please consider storing the configuration as an immutable shared snapshot, for example QSharedPointer<const QDltFilterList>. Update the snapshot only when the configuration changes, then copy only the shared pointer while holding queueMutex before performing the match. This preserves a consistent configuration per message while removing the per-message deep-copy and allocation overhead.

shubhamshahaBMW (changes requested):

shubhamshahaBMW (approved):

shubhamshahaBMW:
Testing was done by @Renu-Priya411. and the review comments are addressed. merging the PR

PR #824, opened 2026-07-24
Allow plugins to enable multiple file/directory selection

lutzbichler:
Some plugins might need informations from multiple input files or directories. These can override the `allowsMultipleFiles` method and return `true` to enable multiple file/directory selection.

Note: While selecting multiple files is possible in a single selection step, selecting multiple directories can only be done by subsequent selection steps.

### DIFF
--- a/qdlt/plugininterface.h
+++ b/qdlt/plugininterface.h
@@ -28,7 +28,7 @@
 
 #include <QString>
 
-#define PLUGIN_INTERFACE_VERSION "1.0.1"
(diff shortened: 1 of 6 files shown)


mfukar:
Hey @lutzbichler  - which plugin would use this functionality?

lutzbichler:
> Hey @lutzbichler - which plugin would use this functionality?

Hi @mfukar,
The (new) SOME/IP plugin. It shall allow to also use Franca[++] as inputs.

PR #825, opened 2026-07-28
Phase 3: search delink with stable snapshot and threaded find-all

Renu-Priya411:
Phase 3: search delink with stable snapshot and threaded find-all


src/searchdialog.cpp
```diff
@@ -312,24 +349,21 @@ void SearchDialog::startParallelFindAll(QRegularExpression searchTextRegExp)
...
             if (buf.isEmpty())
                 continue;
 
             msg.setMsg(buf);
-            msg.setIndex(msgIndex);
+            msg.setIndex(row.messageIndex);
 
             if (doDecode && pluginPtr)
-                pluginPtr->decodeMsg(msg, dlg ? dlg->fSilentMode : 0);
+                (void)pluginPtr->decodeMsgTry(msg, dlg ? dlg->fSilentMode : 0);
```
Pavithra-aa-Anand: decodeMsgTry returns false if tryLock fails.
In parallel Find-All, that return value is ignored, and matching continues. when lock is busy, some messages are searched without decode, so decode-dependent payload matches can be missed. This behaviour is intermittent.

src/searchdialog.cpp
Pavithra-aa-Anand: SearchDialog destructor waits on waitForFinished() after cancellation. If search already completed, there is no issue. The risk appears when closing while Find-All is still running: worker tasks in long I/O or plugin decode can delay completion, so close may feel like an app hang. This is most likely on large logs.

src/dltfileindexer.cpp
```diff
@@ -787,15 +806,29 @@ void DltFileIndexer::run()
     // indexFilter
     if(mode == modeIndexAndFilter || mode == modeFilter)
     {
-        QStringList filenames;
-        for(int num=0;num<dltFile->getNumberOfFiles();num++)
-            filenames.append(dltFile->getFileName(num));
-        if((mode != modeNone) && !indexFilter(filenames))
+        const bool effectiveFilteringEnabled = filtersEnabled && hasActivePositiveOrNegativeFilters(dltFile->getFilterList());
```
Pavithra-aa-Anand: Opened log can appear empty when filters are enabled but no active filter rules exist.
reloadLogFileFinishFilter restores filter mode from filtersEnabled instead of the indexer’s effective filtering result. With settings enabled but no active positive/negative rules, this can re-enable filtering against an empty filter index and make opened logs appear empty.

src/searchdialog.cpp
Renu-Priya411: fixed

src/searchdialog.cpp
Renu-Priya411: fixed

src/dltfileindexer.cpp
Renu-Priya411: fixed

PR #826, opened 2026-08-11
Phase2 live log delink

Renu-Priya411:
Phase 2: live logging delink with index worker and batched UI updates

### DIFF
--- a/qdlt/qdltfile.cpp
+++ b/qdlt/qdltfile.cpp
@@ -778,11 +778,9 @@ bool QDltFile::getMsg(int index,QDltMsg &msg)
     }
 
     // load message from DLT file
-    QByteArray data = getMsg(index);
-    if(data.isEmpty())
+    result = getMsgNoCache(index, msg);
+    if(!result)
         return false;
-    result = msg.setMsg(data,true,dltv2Support);
-    msg.setIndex(index);
 
     // store msg in cache
     if(cacheEnable && result)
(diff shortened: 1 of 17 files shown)


qdlt/qdltfile.cpp
```diff
@@ -765,6 +763,93 @@ bool QDltFile::getMsg(int index,QDltMsg &msg)
...
+    const int originalIndex = index;
+    int num = 0;
+
+    if(index < 0)
+    {
+        qDebug() << "getMsg: Index is out of range" << __FILE__ << "line" << __LINE__;
+        return false;
+    }
+
+    for(num = 0; num < files.size(); num++)
```
Pavithra-aa-Anand, line 822: getMsgNoCache() reads mutable file state outside its mutex. Acquire mutexQDlt before traversing files/indexAll and checking the QFile state. The current lock starts after these accesses, so search/live indexing can race reload/clear and retain a deleted QDltFileItem pointer.

src/searchdialog.cpp
```diff
@@ -312,19 +311,24 @@ void SearchDialog::startParallelFindAll(QRegularExpression searchTextRegExp)
...
                 continue;
 
-            msg.setMsg(buf);
-            msg.setIndex(msgIndex);
-
-            if (doDecode && pluginPtr)
-                pluginPtr->decodeMsg(msg, dlg ? dlg->fSilentMode : 0);
+            if (doDecode)
+            {
+                for(int decoderIndex = 0; decoderIndex < decoderPluginsSnapshot.size(); ++decoderIndex)
```
Pavithra-aa-Anand: Before this change, QDltPluginManager::decodeMsg() held pluginListMutex while invoking a decoder. The new code copies raw plugin pointers and calls them directly, removing that serialization. Decoder implementations are stateful. The same plugin can now be called simultaneously by multiple search chunks and the live worker, causing races, corrupted decoded payloads, or crashes. A snapshot also does not keep a plugin alive while it is unloaded.

So, serialize decoder invocation through a manager-owned mutex while still allowing file reads to run in parallel or give each worker an independent decoder instance/state. Do not use raw snapshots across plugin unload/reload without a lifetime barrier.

Renu-Priya411: fixed

qdlt/qdltfile.cpp
Renu-Priya411: fixed

Pavithra-aa-Anand (approved):

shubhamshahaBMW (approved):

PR #827, opened 2026-08-18
Phase 4 Render Cache Delink

Renu-Priya411:
Phase 4 Render Cache Delink


src/tablemodel.cpp
```diff
@@ -78,176 +237,56 @@ TableModel::TableModel(const QString & /*data*/, QObject *parent)
...
+             {
+                 return QVariant();
+             }
+             return entry.displayValues.at(index.column());
+         }
+     }
+
      std::optional<QDltMsg> msg;
-     if (m_cache.exists(index.row())) {
+     if (m_cache.exists(index.row()))
```
shubhamshahaBMW: m_cache and DecodeRenderCacheEntry both are trying cache, both are needed still?

src/searchtablemodel.cpp
```diff
@@ -31,32 +31,186 @@ SearchTableModel::SearchTableModel(const QString &,QObject *parent) :
...
     pluginManager = NULL;
+    m_renderCacheGeneration = 0;
 }
 
 SearchTableModel::~SearchTableModel()
 {
 
 }
 
+QVariant SearchTableModel::buildDisplayValue(int column, unsigned long messageIndex, QDltMsg &msg) const
```
shubhamshahaBMW: Multiple regressions are found in search case:
1. first index is displayed wrong when enabling filter
2. changing the filter color after search does not reflect in search result (apply filter with colors -> search text -> change color of a filter which is present in search results -> only main table is updated with new color, search result still shows old color)
3. changing the settings like timezone does not apply to search results. (open log -> search -> change timezone from settings -> check main table and search table. only main table is updated with new time)

shubhamshahaBMW (changes requested):

src/searchtablemodel.cpp
```diff
@@ -31,32 +31,186 @@ SearchTableModel::SearchTableModel(const QString &,QObject *parent) :
...
+        unsigned long messageIndex = m_searchResultList.at(index.row());
+
+        DecodeRenderCacheEntry entry;
+        const bool hasRenderCacheEntry = m_decodeRenderCache.exists(index.row());
+        if (hasRenderCacheEntry)
+        {
+            entry = m_decodeRenderCache.get(index.row());
+        }
+
+        const bool isRenderCacheValid = hasRenderCacheEntry
```
Pavithra-aa-Anand: Invalidate the search-model cache when display or marker settings change so existing search results immediately reflect the current configuration.

src/tablemodel.cpp
Renu-Priya411: The caches serve different purposes and are both currently used. They are not duplicates. m_cache caches decoded messages, while DecodeRenderCacheEntry caches formatted rendering data. Removing these caches changes the optimisation

src/searchtablemodel.cpp
Renu-Priya411: Couldnt reproduce the issue 1 and 3. 2nd issue is resolved and have committed the changes. Thanks

src/searchtablemodel.cpp
Renu-Priya411: This behaviour noted in master as well. Thanks

shubhamshahaBMW (approved):
changes tested by @Renu-Priya411 . review comments addressed.

PR #828, opened 2026-08-19
Fix CLI plugin configuration and read-only log handling

Pavithra-aa-Anand:
Add special handling for the fibex_path command when used from CLI: ensure the Non Verbose Mode Plugin actually loads its config, and set the plugin item's filename from the command parameter. Change openDltFile to keep CLI/commandline mode read-only (avoid opening output file for append to not interfere with indexer) and always call reloadLogFile when the file is opened.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1140,19 +1140,33 @@ void MainWindow::commandLineExecutePlugin(QString name, QString cmd, QStringList
         exit(-1);
     }
 
-    // Special handling for the non-verbose decoder plugin when used from
-    // the command line:
-    //
-    // The "fibex_path" command only stores the configured path inside the
-    // plugin. To actually load and parse the Fibex data before any
-    // decoding or exporting happens, we need to trigger its loadConfig()
-    // once the command has been processed. Passing an empty filename lets
-    // the plugin use the path set via the command.
     if (plugin->isDecoder()
-            && plugin->name() == QLatin1String("Non Verbose Mode Plugin")
             && cmd.compare(QLatin1String("fibex_path"), Qt::CaseInsensitive) == 0)
     {
-        plugin->loadConfig(QString());
+        const QString configPath = params.isEmpty() ? QString() : params.at(0);
+        // Non Verbose plugin stores the path without parsing it in command(), other decoder plugins already loaded it above.
+        if (plugin->name() == QLatin1String("Non Verbose Mode Plugin")
+                && !plugin->loadConfig(QString()))
+        {
+            QString msg("Error: ");
+            msg.append(name);
+            msg.append(plugin->error());
+            ErrorMessage(QMessageBox::Warning,name, msg);
+            exit(-1);
+        }
+
+        if (configPath.isEmpty())
+            return;
+
+        for(int num = 0; num < project.plugin->topLevelItemCount (); num++)
+        {
+            PluginItem *pluginitem = (PluginItem*)project.plugin->topLevelItem(num);
+            if(pluginitem->getPlugin() == plugin)
+            {
+                pluginitem->setFilename(configPath);
+                break;
+            }
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/mainwindow.cpp
```diff
@@ -1140,19 +1140,36 @@ void MainWindow::commandLineExecutePlugin(QString name, QString cmd, QStringList
...
-            && plugin->name() == QLatin1String("Non Verbose Mode Plugin")
             && cmd.compare(QLatin1String("fibex_path"), Qt::CaseInsensitive) == 0)
     {
-        plugin->loadConfig(QString());
+        const QString configPath = params.isEmpty() ? QString() : params.at(0);
+                if (plugin->name() == QLatin1String("Non Verbose Mode Plugin"))
+                {
+                    // Non Verbose plugin stores the path but doesn't parse it in command();
+                    // other decoder plugins already loaded it above.
+                    if (!plugin->loadConfig(QString()))
```
mfukar: @Pavithra-aa-Anand these two conditions can be combined in a single `if` statement - makes things easier to read

Could you also fix the indentation level please

src/mainwindow.cpp
```diff
@@ -1140,19 +1140,36 @@ void MainWindow::commandLineExecutePlugin(QString name, QString cmd, QStringList
...
+                        QString msg("Error: ");
+                        msg.append(name);
+                        msg.append(plugin->error());
+                        ErrorMessage(QMessageBox::Warning,name, msg);
+                        exit(-1);
+                    }
+                }
+
+                if (!configPath.isEmpty())
+                {
```
mfukar: Turn this into `if (configPath.isEmpty()) return;`. Makes for easier to read code.

src/mainwindow.cpp
```diff
@@ -1485,17 +1502,14 @@ bool MainWindow::openDltFile(QStringList fileNames)
     /* open existing file and append new data */
     outputfile.setFileName(fileNames.last());
     setCurrentFile(fileNames.last());
-    if( true == outputfile.open(QIODevice::WriteOnly|QIODevice::Append) )
+    // CLI mode stays read-only to avoid interfering with the indexer's Windows size query.
```
mfukar, line 1502: Not related to the review but can you point to some information why/how this might happen? interested to read on it

mfukar (changes requested):

src/mainwindow.cpp
```diff
@@ -1485,17 +1502,14 @@ bool MainWindow::openDltFile(QStringList fileNames)
...
         openFileNames = fileNames;
         isDltFileReadOnly = false;
         //qDebug() << "Opening file(s) wo" << outputfile.fileName() << __FILE__ << __LINE__;
-        if(QDltOptManager::getInstance()->isCommandlineMode())
-            // if dlt viewer started as converter or with plugin option load file non multithreaded
-            reloadLogFile(false,false);
-        else
-            // normally load log file mutithreaded
-            reloadLogFile();
+        reloadLogFile();
```
Renu-Priya411, line 1509: Previously, CLI mode used:
 
if(QDltOptManager::getInstance()->isCommandlineMode())
            // if dlt viewer started as converter or with plugin option load file non multithreaded
            reloadLogFile(false,false);

The new call uses the default behavior, which is normally intended for the GUI. It can run indexing asynchronously, so the CLI may start exporting before the log has been fully indexed. This can result in incomplete or empty conversion output.

The safer implementation would be if(CommandLineMode) - > reloadLogFile(false,false).

Renu-Priya411 (changes requested):

src/mainwindow.cpp
Pavithra-aa-Anand: In normal GUI mode, outputfile is opened for append while DltFileIndexer separately opens or queries the same file during indexing. On Windows, Qt may create the append handle with restrictive sharing flags, which can cause the indexer’s file access or size query to fail or return inconsistent results. Opening the file read-only in CLI mode reduces the requested access and avoids this potential sharing conflict. The exact behavior depends on Qt’s Windows file-sharing implementation and should be verified using the reported Qt/Windows error or Process Monitor.

src/mainwindow.cpp
Pavithra-aa-Anand: added

src/mainwindow.cpp
Pavithra-aa-Anand: if (CommandLineMode) is already handled in the read-only branch at line 1521 of openDltFile(), where CLI mode calls reloadLogFile(false, false). Therefore, it was not duplicated in the surrounding condition. GUI mode continues to use the default asynchronous behavior.

src/mainwindow.cpp
Pavithra-aa-Anand: Indentation is fixed and combined the two checks into a single if condition

Renu-Priya411 (approved):

Renu-Priya411 (approved):

PR #830, opened 2026-09-01
Unload plugins before QApplication shutdown

Pavithra-aa-Anand:
- Add lifecycle management for QPluginLoader instances to ensure plugins are unloaded before QApplication teardown.
- Implement QDltPluginManager::~QDltPluginManager() to unload and delete loaders and to qDeleteAll() plugin objects to avoid crashes during Qt object destruction.

### DIFF
--- a/qdlt/qdltpluginmanager.cpp
+++ b/qdlt/qdltpluginmanager.cpp
@@ -8,6 +8,8 @@
 #include <QMutex>
 #include <QTextStream>
 #include <QString>
+#include <QtAlgorithms>
+#include <utility>
 
 #ifndef PLUGIN_INSTALLATION_PATH
 #define PLUGIN_INSTALLATION_PATH ""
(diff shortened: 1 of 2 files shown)


qdlt/qdltpluginmanager.cpp
```diff
@@ -18,6 +20,21 @@ int QDltPluginManager::size() const
     return plugins.size();
 }
 
+QDltPluginManager::~QDltPluginManager()
+{
+    // Unload plugins before QApplication shuts down; otherwise the OS unloads them too late and Qt GUI objects may crash during destruction.
+    qDeleteAll(plugins);
+    plugins.clear();
```
Renu-Priya411, line 28: qDeleteAll and plugins.clear() does do different tasks. Add a comment of what each does. To just see seems as though both are doing the same clearing task of plugins

Renu-Priya411 (changes requested):

qdlt/qdltpluginmanager.cpp
Pavithra-aa-Anand: Added single line comments

Renu-Priya411 (approved):

PR #834, opened 2026-09-03
Fix live view not updating when connecting a single ECU

### LINKED ISSUE #833: Live view and Recv counters never update when connecting a single ECU (regression in 2.30.0)
### Describe the bug

Connecting a **single ECU** (Config → Connect, the ECU context menu, or the toolbar button with an ECU selected) leaves the GUI completely static:

* the message table stays empty,
* the `Recv:` / `Recv Errors:` / `Sync found:` status bar counters stay at `0`.

Data *is* received correctly the whole time — the socket is `ESTABLISHED`, the receive queue drains, and the messages are parsed and written to the log file. Only the GUI never refreshes. Reconnecting, toggling filters or changing `autoScroll` has no effect.

Using **Connect All** instead works fine, which makes the bug look like an ECU/network problem rather than a GUI one. In my case I spent a long time debugging the target and the network before realising the transport was never at fault.

### Root cause

`drawUpdatedView()` is what refreshes both the table (`tableModel->modelChanged()`) and the status bar counters. It is driven by `drawTimer`, and `drawTimer` is started in exactly one place — `connectAll()`:

https://github.com/COVESA/dlt-viewer/blob/master/src/mainwindow.cpp#L4086-L4091

```cpp
void MainWindow::connectAll()
{
    ...
    // periodically update table view to account for the new incoming messages
    const int drawInterval = (settings->RefreshRate > 0) ? 1000 / settings->RefreshRate
                                                         : 1000 / DEFAULT_REFRESH_RATE;
    drawTimer.start(drawInterval);
    connect(&drawTimer, &QTimer::timeout, this, &MainWindow::drawUpdatedView);
}
```

`connectECU()` — the code path used for a single-ECU connect — never starts the timer. `grep -n drawTimer src/mainwindow.cpp` returns only the start in `connectAll()`, the stop in `disconnectAll()` and the member declaration.

This is a regression from 6f0c092 ("Move autoscroll timer to appropriate place"), which removed the data-driven single-shot kick that used to live at the end of `updateIndex()`:

```cpp
    if (!draw_timer.isActive()) {
        const int drawInterval = ...;
        draw_timer.start(drawInterval);
    }
```

Because that kick was driven by *arriving data*, it was connection-path agnostic and the single-ECU case worked. Moving the start into `connectAll()` silently dropped the `connectECU()` path. The later follow-up 06c1590 ("Fix live table updates when autoscroll is off") removed the `autoScroll` guard but kept the start in `connectAll()` only, so the single-ECU case is still broken.

First released in **2.30.0** (`git describe --contains 6f0c092` → `2.30.0~39^2`); still present on `master` (edf7aa5).

Two smaller issues in the same lines:

1. `connect(&drawTimer, ...)` is inside `connectAll()`, so every *Connect All* adds another connection to the same signal. After N Connect All cycles `drawUpdatedView()` runs N times per tick.
2. In the constructor, the `autoConnect` block calls `connectAll()` *before* `connect(&timer, ...)` is executed, so a timer started via auto connect has no receiver wired up yet.

### To Reproduce

1. Create a project with a **single** ECU (TCP) that produces a continuous log stream.
2. Make sure `autoConnect` is off.
3. Select the ECU and use **Config → Connect** (not *Connect All*).
4. The ECU shows as connected and the log file on disk grows, but the table stays empty and `Recv:` stays `0`.

### Expected behaviour

The live view and the status bar counters update regardless of whether the connection was initiated via *Connect All* or a single-ECU connect.

### Evidence

Measured against a real, continuously logging ECU, with a probe `qDebug()` added at the top of `drawUpdatedView()`, `QT_QPA_PLATFORM=offscreen`, 30 s per run, identical isolated `HOME` and project for both builds:

| single-ECU connect (`autoConnect=0`) | `drawUpdatedView()` calls in 30 s | rows in table |
| --- | --- | --- |
| `master` (edf7aa5) | **0** | – |
| with the fix | **560** | 10392 |

Meanwhile the receive path was provably healthy on `master`: the socket stayed `ESTABLISHED` with a drained receive queue and the viewer's cache file grew steadily (~120 KB every 3 s, 6.6 MB total) while the GUI showed nothing.

`Connect All` is unaffected by the fix (498 vs 497 refreshes in 25 s).

### Environment

* DLT Viewer: 2.30.0 and `master` (edf7aa5)
* Qt 5, Ubuntu 24.04

I have a fix ready and will open a PR referencing this issue.

stensson:
Fixes #833.

## Problem

`drawUpdatedView()` refreshes the message table and the `Recv:` / `Recv Errors:` / `Sync found:` status labels. It is driven by `drawTimer`, which was started in exactly one place — `connectAll()`.

Connecting a **single ECU** goes through `connectECU()` (Config → Connect, the ECU context menu, or the toolbar button with an ECU selected), which never started the timer. The result is a completely static GUI: the table stays empty and the counters stay at `0`, even though data is received and written to the log file correctly the whole time.

This is a regression from 6f0c092, which replaced the data-driven single-shot kick at the end of `updateIndex()` with a start in `connectAll()`. The old kick was triggered by arriving data and was therefore connection-path agnostic. Released in 2.30.0.

## Change

The timer start/stop moves into a small helper, `updateDrawTimerState()`, which keys off the existing `isLiveLoggingActive()` predicate and is called from `connectECU()` and `disconnectECU()`. Since `connectAll()`/`disconnectAll()` delegate to those, every connect/disconnect path is now covered, and the timer runs exactly while at least one ECU is live. Stopping also performs a final `drawUpdatedView()` so messages that arrived between the last tick and the disconnect are rendered rather than silently missing from the view.

Two related fixes in the same lines:

* `connect(&drawTimer, ...)` moves to the constructor. Previously it ran on every *Connect All*, so each cycle added another connection to the same signal and `drawUpdatedView()` was invoked N times per tick after N connects.
* The `autoConnect` block in the constructor moves *below* the timer wiring. It called `connectAll()` before `connect(&timer, ...)` had run, so timers started via auto connect had no receiver connected yet.

No behaviour change for the *Connect All* path.

## Verification

Tested against a real ECU producing a continuous DLT stream. A probe `qDebug()` was added at the top of `drawUpdatedView()` in **both** builds, run with `QT_QPA_PLATFORM=offscreen` and an identical isolated `HOME` and project file.

Single-ECU connect (`autoConnect=0`), 30 s per run:

| build | `drawUpdatedView()` calls | rows in table |
| --- | --- | --- |
| `master` (edf7aa5) | **0** | – |
| this PR | **560** | 10392 |

On `master` the receive path was provably healthy at the same time — socket `ESTABLISHED` with a drained receive queue, and the viewer's cache file growing ~120 KB every 3 s (6.6 MB total) — while the GUI showed nothing.

Regression check, *Connect All* with `autoConnect=1`, 25 s per run:

| build | `drawUpdatedView()` calls |
| --- | --- |
| `master` | 498 |
| this PR | 497 |

Builds clean with Qt 5 on Ubuntu 24.04. The probe instrumentation is **not** part of this PR.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -180,14 +180,19 @@ MainWindow::MainWindow(QWidget *parent) :
         exit(0);
     }
 
+    /* start timer for autoconnect */
+    connect(&timer, SIGNAL(timeout()), this, SLOT(timeout())); // we want to start the timer only when an ECU connection is active
+
+    /* periodic refresh of the live view; started only while an ECU connection is active */
+    connect(&drawTimer, &QTimer::timeout, this, &MainWindow::drawUpdatedView);
+
     /* auto connect */
+    /* must happen after the timers above are wired up, otherwise the timers
+     * started by connectAll() would have no receiver connected yet */
     if( (settings->autoConnect != 0) ) // in convertion mode we do not need any connection ...)
     {
         connectAll();
     }
-
-    /* start timer for autoconnect */
-    connect(&timer, SIGNAL(timeout()), this, SLOT(timeout())); // we want to start the timer only when an ECU connection is active
 
     restoreGeometry(settings->geometry);
     restoreState(settings->windowState);
@@ -4083,17 +4088,10 @@ void MainWindow::connectAll()
         EcuItem *ecuitem = (EcuItem*)project.ecu->topLevelItem(num);
         connectECU(ecuitem);
     }
-
-    // periodically update table view to account for the new incoming messages
-    const int drawInterval = (settings->RefreshRate > 0) ? 1000 / settings->RefreshRate
-                                                         : 1000 / DEFAULT_REFRESH_RATE;
-    drawTimer.start(drawInterval);
-    connect(&drawTimer, &QTimer::timeout, this, &MainWindow::drawUpdatedView);
 }
 
 void MainWindow::disconnectAll()
 {
-    drawTimer.stop();
     for(int num = 0; num < project.ecu->topLevelItemCount (); num++)
     {
         EcuItem *ecuitem = (EcuItem*)project.ecu->topLevelItem(num);
@@ -4105,6 +4103,7 @@ void MainWindow::disconnectAll()
     if(settings && settings->includeManualMarkersInFilter && !isLiveLoggingActive())
         updateManualMarkerUnionInFilter();
 
+    updateDrawTimerState();
     checkConnectionState();
 }
 
@@ -4140,6 +4139,8 @@ void MainWindow::disconnectECU(EcuItem *ecuitem)
     // If this was the last active ECU, switch back to offline marker union (if enabled).
     if(settings && settings->includeManualMarkersInFilter && !isLiveLoggingActive())
         updateManualMarkerUnionInFilter();
+
+    updateDrawTimerState();
 }
 
 void MainWindow::on_action_menuConfig_Connect_triggered()
@@ -4204,6 +4205,8 @@ void MainWindow::connectECU(EcuItem* ecuitem,bool force)
         ecuitem->connected = false;
         ecuitem->update();
         on_configWidget_itemSelectionChanged();
+
+        updateDrawTimerState();
 
         /* reset receive buffer */
         ecuitem->totalBytesRcvd = 0;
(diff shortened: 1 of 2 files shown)


stensson:
Friendly ping for a maintainer: as this is my first contribution here, the CI run is sitting in `action_required` and needs a manual "Approve and run workflows" before the builds will start. Thanks!

stensson:
Closing this PR — the underlying bug has since been fixed upstream by an unrelated refactor of the live-view refresh path (commits ed403cb "Phase2 live log delink" #826, ee0fc88 "Phase 4 Render Cache Delink" #827, 58c3e40 "Phase 5: filter worker..." #823).

`master` no longer drives the live table/`Recv:` counters from `drawTimer`, started only in `connectAll()`. It now uses an async `liveIndexWorker` + `liveBatchTimer` → `applyLiveBatchUpdate()`, wired up once in `initFileHandling()` independent of which/how many ECUs are connected — so the single-ECU-connect case that used to leave the GUI frozen now works correctly.

Re-verified against a real ECU with a continuous DLT stream, on clean unpatched `master` (`ed403cb`):

| scenario | `applyLiveBatchUpdate()` calls | rows |
| --- | --- | --- |
| single-ECU connect, autoConnect off, 30s | **854** | 9351 → 9387 |
| Connect All, autoConnect on, 25s | 755 | → 19142 |

Both paths now update the table and status bar counters live, so the fix in this PR is no longer needed and would just reintroduce a redundant/dead `drawTimer` mechanism alongside the new one. Closing without merging.

Thanks for the refactor that resolved this as a side effect!

